PR #3, opened 2021-04-29
ToC for pages at the top. probably needs some spicing up via css

civviGH:


civviGH:
i did not check the tests yet. maybe the refactor of the formatter.py conflicts with sth there

redimp:
Outdated, since the TOC is now a part of the page.

PR #34, opened 2022-10-15
Feature/warn non version control

marcrleonard:
This will show a warning if the .md file is on disk but is not in version control


redimp:
Thanks for this feature .. with test! Awseome! (But I have to dig into the test system now and find out what is broken there)

marcrleonard:
Oh, shoot, did I break something? I thought I made sure all tests passed before I pushed...

redimp:
You broke nothing, the bug is somewhere in the test system (see #36) Your commit and your test is fine!

PR #42, opened 2022-10-20
Feature/subdirectories

marcrleonard:
Here it is! Give it a try! I've included tests for for sub directories, as well as a separate rename test. There are a few other minor improvements like a fix for an incorrect redirect function call and some incorrect closing HTML tags.


redimp:
Merged in 481f2add6fa16abda60ea7c9a2593116367dd1a8.

PR #88, opened 2024-02-21
Fixes #85 - adds support for case sensitive page names

### LINKED ISSUE #85: Page names case not reflected in filenames
Right now, page names are normalized to lowercase in Git and automatically raised to title case in the UI. The user should have the option to preserve case, so wikis written with other systems (that don't have all lowercase page names) can be imported seamlessly.

Related to #61.

weaversam8:
See discussion in #85. This adds the ability for page names to resolve case-sensitively, but only when the `RAW_PAGE_NAMES` configuration option is set.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -36,7 +36,7 @@ for EV in READ_ACCESS WRITE_ACCESS ATTACHMENT_ACCESS; do
         echo "${EV} = '${!EV}'" >> ${OTTERWIKI_SETTINGS}
     fi
 done
-for EV in AUTO_APPROVAL EMAIL_NEEDS_CONFIRMATION; do
+for EV in AUTO_APPROVAL EMAIL_NEEDS_CONFIRMATION RAW_PAGE_NAMES; do
     if [ ! -z "${!EV}" ]; then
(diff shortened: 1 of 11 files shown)


weaversam8:
Open to suggestions here - I do think that passing a `raw_page_names` argument into those utility functions all over the place does get a bit inconvenient, but I'm not sure of a better way to do it without creating a circular import.

redimp:
Thank you for the work! From the first read through I have to say it looks good.

Will look into the details on Friday (unfortunately got a full day today and tomorrow).

redimp:
I have adopted your commits and worked a little on them. Please double check that the branch [feature/retain-page-name-case](https://github.com/redimp/otterwiki/tree/feature/retain-page-name-case) does what you want and solves $85 for you.

Details: With moving the functions `get_filename`, `get_attachment_directoryname` and `get_pagename` into `otterwiki.helper` they can access `app.config` and don't need the extra argument anymore.

I've renamed the config setting to `RETAIN_PAGE_NAME_CASE` .. this sounds less "raw" to me.

redimp:
Thanks for double checking and for providing the patch.

Will add this to the next release (this week).

(I forgot to send this yesterday)

PR #95, opened 2024-03-06
Proof of concept for reverse-proxy auth

### LINKED ISSUE #90: Support auth via reverse-proxy headers
OtterWiki currently authenticates users based on the sqlite database it maintains. If hosted in a corporate environment, it may be beneficial to install a reverse-proxy that handles authentication (with SSO, etc...) and then grants the user access to the wiki.

This is technically possible today, but you lost the semantics of tracking which user is editing the wiki. I'd like to propose a configuration option that would allow OtterWiki to use values provided by trusted HTTP headers instead. Maybe something like:

- `X-OtterWiki-Email` - The email address of the user
- `X-OtterWiki-Username` - The "name" of the user attached to the commit
- `X-OtterWiki-Role` - The role of the user, to control access.

weaversam8:
- This commit adds a new auth manager class for authorizing via proxy headers `ProxyHeaderAuth` which can be selected by setting the `AUTH_METHOD` env var to `PROXY_HEADER`

    - This auth manager looks for the following headers in order to
      create a "pseudo-user" on each request. No users are committed to
      the SQLite database when using this auth manager.

        - `X-OtterWiki-Name` - the name of the user to include on the Git commit when editing a page
        - `X-OtterWiki-Email` - the email of the user to include on the Git commit when editing a page
        - `X-OtterWiki-Permissions` - a comma separated list of permissions to grant to the user

    - The Docker `entrypoint.sh` script has been updated to pass the `AUTH_METHOD` config option thru if set in the environment.

- `has_permission(permission, user)` is now a method specific to each auth manager

- auth managers now implement a `supported_features()` method to detail which features they support (like whether an auth manager allows a user to change their name or password, or logout)

    - the features object this method returns is present in all Jinja
      templates as the variable `auth_supported_features`

    - the settings page was updated to prevent a user from editing their
      password and name if it is not supported by the current auth
      manager.

    - the dropdown menu present on all page was updated to hide the
      "logout" button if it is not supported by the current auth manager

- The `test_settings.py` test was updated to tolerate extra whitespace

CC @redimp. I didn't make the headers configurable for now, and I'm sure this is a bit hacky, so interested to hear what you think.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -30,7 +30,7 @@ for EV in SITE_NAME SITE_LOGO SITE_DESCRIPTION SITE_ICON; do
     fi
 done
 # permissions
-for EV in READ_ACCESS WRITE_ACCESS ATTACHMENT_ACCESS; do
+for EV in AUTH_METHOD READ_ACCESS WRITE_ACCESS ATTACHMENT_ACCESS; do
     if [ ! -z "${!EV}" ]; then
         sed -i "/^${EV}.*/d" ${OTTERWIKI_SETTINGS}
         echo "${EV} = '${!EV}'" >> ${OTTERWIKI_SETTINGS}
--- a/docs/auth_examples/header-auth/Caddyfile
+++ b/docs/auth_examples/header-auth/Caddyfile
@@ -1,0 +1,17 @@
+{
+	http_port 8081
+	auto_https off
+	log {
+		format console
+	}
+	servers {
+		log_credentials
+	}
+}
+:8081 {
+	reverse_proxy otterwiki:80 {
+		header_up x-otterwiki-name "Otter Example"
+		header_up x-otterwiki-email "mail@example.com"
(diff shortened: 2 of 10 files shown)


redimp:
That looks very good. I don't have time to set up a test environment until the weekend. I will then test and merge or let you know if I run into problems.

Awesome work @weaversam8 !

redimp:
First test went fine!

I have added the commits in the branch [feature/proxy-header-auth](/redimp/otterwiki/tree/feature/proxy-header-auth).

I'm closing this pull request, but this in no way means that it has been rejected.

weaversam8:
I just pulled and tested these changes. Fixed one tiny bug in 7b3098c but otherwise this looks great! I think this example is plenty sufficient to show how this is used, thanks for putting that together!

I rebased my branch off of main and pushed, so this PR should be good to merge (since it now contains your changes.)

redimp:
Merged. Thank you for the work.

Will do a new release including this feature within the next two days.

PR #96, opened 2024-03-06
Plugin example: move wikilink processing

weaversam8:
- This PR uses the `pluggy` framework mentioned in #91 to move the "wikilink" parsing functionality to a separate "plugin" that is still housed within the OtterWiki codebase.

- The wikilink parsing code is unchanged, and was just moved to a WikiLinkPlugin class, which implements the new `preprocess_markdown` hook.

- The WikiLinkPlugin is manually registered, but this commit also includes code to automatically discover plugins installed as Python packages if they use the "otterwiki" Setuptools entrypoint. This has yet to be tested, but should work.

- Tests have been updated.

CC @redimp - what do you think? This should be a really easy/simple hook to start with.

### DIFF
--- a/otterwiki/otterwiki_plugins.py
+++ b/otterwiki/otterwiki_plugins.py
@@ -1,0 +1,79 @@
+# This file contains both the "plugin spec" for OtterWiki, and several
+# plugins included with the app itself.
+
+import pluggy
+import re
+import urllib
+
+hookspec = pluggy.HookspecMarker("otterwiki")
+hookimpl = pluggy.HookimplMarker("otterwiki")
+
+
+class OtterWikiPluginSpec:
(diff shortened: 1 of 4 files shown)


redimp:
I'm running out of superlatives. This is brilliant.

`preprocess_markdown` is a well picked as proof of concept. 

TBH: Using preprocessing for handling the wiki links was a workaround, because at this time my approaches to extend mistune were not working as I expected. Happy to see that this somehow became useful. 😅

Still need to understand how pluggy works exactly, but that will be easy with this PR. Will merge as soon as I get the hang of it.

I'll check tomorrow or at the weekend to see what other places there are that are suitable for hooks. I look forward to your suggestions.

PR #98, opened 2024-03-08
Allow disabling of registration

Hellrespawn:
Adds an option to disable registration.

When enabled, this will hide the link to register on the login and lost password page. Manual GET and POST requests to `/register` will redirect to `index` with an error message in a toast. 

I'm not familiar with Flask, so I don't know if there is a more idiomatic way of handling an invalid POST-request.

It does not touch the link in `initial_home.md`, but registration is enabled by default, so this shouldn't matter.


redimp:
Thanks for the PR!

Will look into the details tomorrow. Had this on my personal wishlist. I may also add the ability for admins to invite users/create user accounts.

PR #143, opened 2024-09-24
Podman and SELinux fixes to the Docker setup

mattock:
The current Docker / Docker Compose setup does not work on Podman due to volume mount issues. It also does not work if SELinux is in *enforcing* mode, because volume mounts are not correctly labeled.

After applying these changes the setup will work with rootless Podman and SELinux. The changes have been tested on Rocky Linux 9.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ COPY pyproject.toml MANIFEST.in README.md LICENSE /src/
 WORKDIR /src
 
 # install requirements
-RUN --mount=type=cache,target=/root/.cache \
+RUN --mount=type=cache,target=/root/.cache,Z \
     python -c 'import toml; print("\n".join(toml.load("./pyproject.toml")["project"]["dependencies"]));' > requirements.txt && \
     pip install -r requirements.txt
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,8 @@
 # to get started.
 #
 version: '3'
+volumes:
+  otterwiki:
 services:
   web:
     build: .
@@ -17,4 +19,4 @@ services:
     volumes:
       # mount the local directory ./app-data into the container
       # (so settings, database and repository are accessible)
-      - ./app-data:/app-data
+      - otterwiki:/app-data:rw,z


docker-compose.yml
```diff
@@ -17,4 +19,4 @@ services:
     volumes:
       # mount the local directory ./app-data into the container
       # (so settings, database and repository are accessible)
-      - ./app-data:/app-data
+      - otterwiki:/app-data:rw,z
```
redimp, line 22: With this change not a local directory would be mounted, but a volume created and stored somewhere in the docker environment, e.g. `/var/lib/docker/volumes/...`, depending on the volume driver.

There is nothing to be said against this, but it must be correctly commented.

In the case that a user want to mount the local directory and use `SELINUX=enforcing` a change of the directories context via `semanage` might be the way to go.

Dockerfile
```diff
@@ -21,7 +21,7 @@ COPY pyproject.toml MANIFEST.in README.md LICENSE /src/
 WORKDIR /src
 
 # install requirements
-RUN --mount=type=cache,target=/root/.cache \
+RUN --mount=type=cache,target=/root/.cache,Z \
```
redimp, line 24: Tested this with `Docker version 27.2.0, build 3ab4256` on MacOS and received a:

```
Dockerfile:24
--------------------
  23 |     # install requirements
  24 | >>> RUN --mount=type=cache,target=/root/.cache,Z \
  25 | >>>     python -c 'import toml; print("\n".join(toml.load("./pyproject.toml")["project"]["dependencies"]));' > requirements.txt && \
  26 | >>>     pip install -r requirements.txt
  27 |
--------------------
ERROR: failed to solve: invalid field 'Z' must be a key=value pair
```

I was not able to find the `Z` syntax in the docker documentation, can you please point me to it?

redimp:
I will set up a VM with Rocky Linux 9 and `SELINUX=enforcing` and dig into this.

Dockerfile
redimp: Found the documentation for `:Z` https://docs.docker.com/engine/storage/bind-mounts/#configure-the-selinux-label .

For the sake of maximum compatibilty maybe I should just drop the `mount=type=cache` from the Dockerfile? I added this when I was on the road with poor internet.

docker-compose.yml
redimp: I was able to run podman with a mount from a local directory using

```
podman run --rm -p 8080:80 -v $PWD/app-data:/app-data:Z redimp/otterwiki:2
```

I think the documentation needs an update on running An Otter Wiki in an enviroment with `SELinux=enforcing`.

@mattock What is the current best way to get `docker compose` (or `podman compose`?) running on Rocky9?

mattock: > With this change not a local directory would be mounted, but a volume created and stored somewhere in the docker environment, e.g. `/var/lib/docker/volumes/...`, depending on the volume driver.

Yep. The location when using podman depends on whether you're running podman/podman-compose as root or as a normal user. In the former case volume(s) end up under /var/lib/<something> like with Docker, whereas in the latter case they are placed under ~/.local/share/containers/storage/volumes/.

> There is nothing to be said against this, but it must be correctly commented.

Fully agreed.

> 
> In the case that a user want to mount the local directory and use `SELINUX=enforcing` a change of the directories context via `semanage` might be the way to go.

Semanage + restorecon + chown + chmod should work. I'm not 100% sure if just adding the "z" o "Z" volume mount option could also resolve this. In general I find using the default volume locations easier with Podman, but that's just my personal preferences.

mattock:
> I will set up a VM with Rocky Linux 9 and `SELINUX=enforcing` and dig into this.

Great! Let me know if you want me to test out something - I'm more than happy to do it. The more I can upstream the better :smile:.

redimp:
I've tested deploying An Otter Wiki on Rocky9 with `SELINUX=enforcing` with both `podman` and `docker` it worked well. For podman the `:z` flag for the bind-mount was required, docker seems to set the required flags per default.

I covered this in the bran dnew [FAQ](https://otterwiki.com/FAQ#environments-with-selinux) and added a info box to the podman part of the [Install Documentation](https://otterwiki.com/Installation#podman-and-podman-compose).

Thank yor for bringing this to my attention!

PR #148, opened 2024-10-10
Page tree colors

ervinBLU:
Page tree colors, when having many pages in a category and subcategories with its pages, it takes a while to find what you look for. Having categories and subcategories in green text and pages in blue, headlines in red is much more pleasant to find what you look for.

I used the color utilities from Halfmoon

Categories: .text-success
Pages: .text-primary
Headers: .text-danger

example image

image

image

### DIFF
--- a/otterwiki/templates/pageindex.html
+++ b/otterwiki/templates/pageindex.html
@@ -17,12 +17,16 @@
 <div class="pageindex-letterblock"><h2 class="content-title font-weight-bolder pageindex-header">{{letter}}</h2>
 {% for depth, title, url, page_toc in pagelist %}
 <div class="pageindex-pageblock" style="padding-left:{{1.5 * depth}}rem;">
-<a href="{{ url }}" class="global-toc-link font-weight-semi-bold">{{title}}</a>
+{% if '/' in title %}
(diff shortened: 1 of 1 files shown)


redimp:
Hey @ervinBLU, thank you for your pull request. I'm personally not good with colors, they distract me and I'm never happy with any combination.

I agree that the Page Index could need some improvedment to make Folders, Pages and Headings more distinguishable, but this amount of colors is in my opinion too much. What do you think of these modifications, with bold folders and italic headings?

image

What do you think?

To make this customizeable, I will add classes to the different types of items, so that they can be changed with a [custom css](https://otterwiki.com/Customization#customcss):

- `pageindex-header`
- `pageindex-folder`
- `pageindex-page`
- `pageindex-heading`

ervinBLU:
Yea I understand, in a way having lot's of colors is a bit distracting, your solution is fine as well, but if there is a way to customize them with .css then yea, that would be perfect. Thank you.

redimp:
The proposed compromise (see [b678033](https://github.com/redimp/otterwiki/commit/b678033faed3a9344619226604dbfc8f2300d27e) ) bas been released in [v2.7.0](https://github.com/redimp/otterwiki/releases/tag/v2.7.0).

PR #162, opened 2024-11-07
Draft: add production docker-compose setup

mattock:
This a draft or "proof of concept" for a "production" Otterwiki than can be run with Docker Compose or Podman Compose. This is meant to be a basis for discussion and not a finished product.

This Otterwiki setup does not make use of supervisord. It reuses the work done in Dockerfile-slim with a minor change (http-socket -> socket) required by the nginx reverse proxy. The setup is composed of three containers:

* *web:* uwsgi process that serves otterwiki
* *nginx:* reverse proxy for uwsgi
* *db:* runs the postgresql database for otterwiki

These containers are configured with environment files:
* .env.prod.web
* .env.prod.nginx
* .env.prod.db

There are four volumes:
* *db*: used by the *db* container
* *app-data*: configuration files for the *web* (otterwiki/uwsgi) container
* *static*: shared static data for *web* (rw) and *nginx* (ro) containers
* *letsencrypt*: certificates used by *nginx* (ro) container

The assumption is that nginx will use Letsencrypt certificates. Creating an renewing those certificates is done outside of the Compose setup. A certbot renewal script for Podman is provided [nginx/certbot-renew.sh](nginx/certbot-renew.sh).

Despite the naming the certificate setup is not tied with Letsencrypt. Renaming the volume and the paths would make it the SSL certificate setup more generic and less confusing when Letsencrypt is not used. 

This code has only been tested with Podman on Rocky Linux 9. Theoretically it should work equally well on Docker, except for the certbot-renew.sh script which would require a few minor changes.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,0 +1,40 @@
+---
+version: '3'
+volumes:
+  app-data:
+  letsencrypt:
+  static:
+  db:
+services:
+  db:
+    image: postgres:16.4
+    restart: unless-stopped
+    volumes:
+      - db:/var/lib/postgresql/data:rw,z
+    env_file: .env.prod.db
+  nginx:
+    build:
+      context: nginx
+    restart: unless-stopped
+    env_file:
+      - .env.prod.nginx
+    volumes:
+      - letsencrypt:/etc/letsencrypt:ro,z
+      - static:/app/otterwiki/static:ro,z
+    ports:
+      - "8443:443"
+  web:
+    build:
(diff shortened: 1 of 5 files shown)


mattock:
@redimp does this look somewhat reasonable to you?

redimp:
Sorry for the late review.

Thinking about this: I'm not sure if this should be part of the repo or the image, but it would be a useful additon as tutorial to the documentation. With. additional remarks for handling SELinux, podman and the certbot.

mattock:
Yeah, no worries. I'll deploy the thing and send documentation your way.

Issue #54, opened 2023-11-03
[Feature Request] Allow admins to set pages as 'protected', allowing only Admins to edit those pages


SeaswimmerTheFsh:
Title explains it, pretty much. This is useful for pages like the home page that you don't really want random people editing, but you want the rest of the wiki to be editable by other people.

redimp:
Thx for opening this feature request. I've thought about makeing pages configureable via YAML frontmatter. For this feature it might look like this:

```markdown
---
protected: true
---

# Home

...
```

Do you that's a viable option?

SeaswimmerTheFsh:
Would there be anything preventing a normal user from adding that to a page and making it un-editable, requiring admin intervention? Or would only admins be allowed to save documents containing that type of YAML configuration?  

Also, if that's going to be a thing, maybe add a hidden option as well, to prevent certain pages from appearing in the search/A-Z list?  
That might be a separate feature request, but otherwise yeah that works.

redimp:
Obviously I underestimate what disruptive or trolling peoples might could come up with. Good point. Configurations like should be limited to Admins.

My assumption was that if someone encounters disruptive behaviour, the user gets the approval revoked and the changes to the pages reverted.

redimp:
> Also, if that's going to be a thing, maybe add a hidden option as well, to prevent certain pages from appearing in the search/A-Z list? That might be a separate feature request, but otherwise yeah that works.

To make this possible, a permissions management per page or per subdirectory would be the most viable option. I would consider this (for now) as out of scope. To discuss this further, please open another feature request. I'm curious what your use-case is.

redimp:
Another thought: Maybe permissions per page / subdirectory would be the best way to handle protected pages as well.

SeaswimmerTheFsh:
> Obviously I underestimate what disruptive or trolling peoples might could come up with. Good point. Configurations like should be limited to Admins.
> My assumption was that if someone encounters disruptive behaviour, the user gets the approval revoked and the changes to the pages reverted.

Yeah, I think that's the general idea for the most part, but for some pages you kinda don't want approved users to edit them either, like a Rules page or a homepage, or high-traffic pages in general.

> To make this possible, a permissions management per page or per subdirectory would be the most viable option. I would consider this (for now) as out of scope. To discuss this further, please open another feature request. I'm curious what your use-case is.

I'll probably open another feature request at some point, but for now it's not really important. Just a random thought I had while responding last night.

[...]

PR #166, opened 2024-11-24
ldap provider

metallkopf:
Add ldap as an external authentication provider.

pending work in no specific order:

* [ ] update documentation
* [ ] include tests
* [ ] database migration
* [ ] whatever i forgot

### DIFF
--- a/otterwiki/auth.py
+++ b/otterwiki/auth.py
@@ -31,6 +31,12 @@ from otterwiki.models import User as UserModel
 from datetime import datetime
 import hmac
 
+try:
+    import ldap
+    has_ldap = True
+except ImportError:
+    has_ldap = False
+
 
 def check_password_hash_backport(pwhash, password):
     # split pwhash to check the method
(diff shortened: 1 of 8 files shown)


redimp:
Hey @metallkopf  thank you for this PR! I will take a look at it, set up a test environment and check whether everything works as intended. Unfortunately I'm currently swamped at work,  will take care of this on the weekend.

metallkopf:
This is what am using:

* to build: podman build -t ldap .
* to run: podman run --rm -it -p 3890:389 ldap
* settings.cfg:
  - LDAP_URI = 'ldap://localhost:3890'
  - LDAP_USERNAME = 'cn=Manager,dc=ldap,dc=local'
  - LDAP_PASSWORD = 'secret'
  - LDAP_BASE = 'dc=ldap,dc=local'
  - LDAP_SCOPE = 'subtree'
  - LDAP_DOMAIN = 'ldap.org'
* user / password:
  - john@ldap.org / 12345678
  - fulano@ldap.org / password
  - max@ldap.org / qwertyui

[container.zip](https://github.com/user-attachments/files/17959525/container.zip)

redimp:
This looks promising! Unfortunately I did not get my example working ... can you check (and maybe fix) the example in [docs/auth_examples/ldap-auth](https://github.com/metallkopf/otterwiki/tree/ldap/docs/auth_examples/ldap-auth)?

Things I want to check when we got this working:

- [ ] Add an option to auto-create users existing in ldap
- [ ] Consider to move `LdapAuth` in a seperate class (like `ProxyHeaderAuth`)

otterwiki/auth.py
```diff
@@ -31,6 +31,13 @@
 from datetime import datetime
 import hmac
 
+try:
+    import ldap
+    has_ldap = True
+except ImportError as e:
+    app.logger.debug(f"Unable to import ldap: {e}")
```
metallkopf, line 38: this try/catch block was added to allow ldap to be an optional dependency

metallkopf:
i couldnt get those docker services to work, so i integrated them into one.

the problem: not all variables were defined in server.py, so they couldnt be read from the environment.

> Add an option to auto-create users existing in ldap

you mean auto-register on login?, sure the common name can be recovered after binding 

> Consider to move LdapAuth in a seperate class (like ProxyHeaderAuth)

i was considering ldap as a secundary login, allowing both type of users. but we can work something out


ps: do you any preference on how to implement database migrations?

redimp:
> i couldnt get those docker services to work, so i integrated them into one.

Looks good to me. Have not tested this yet.

> you mean auto-register on login?

Yes, this should be an option.

> do you any preference on how to implement database migrations?

I'm open for suggestions. I've used [flask-migrate](https://flask-migrate.readthedocs.io/en/latest/) before.

redimp:
Got the example working! Have to think about the best way to integrate this. 

But even with a seperate `LdapAuth` some work has to be done, like fetching information from ldap about the users, e.g. the Name. Or for future features groups ... ?

otterwiki/preferences.py
```diff
@@ -408,6 +408,7 @@ def handle_user_add(form):
     # update user from form
     user.name = form.get("name").strip()  # pyright: ignore
     user.email = form.get("email").strip()  # pyright: ignore
+    user.provider = form.get("provider").strip()  # pyright: ignore
```
metallkopf: this addition was producing an error on my end

metallkopf:
> ...Have to think about the best way to integrate this.

you luck with that

> ...like fetching information from ldap about the users, e.g. the Name. 

i've already added retrieving the -common- name of an user when he self-registers

> ...Or for future features groups ... ?

can you elaborate a bit more

redimp:
> > ...Or for future features groups ... ?

> can you elaborate a bit more

The request for more fine grained permissions exists for a long time. I think, that for a useful implementation a concept groups is necessary (For some more details, please see https://github.com/redimp/otterwiki/issues/54#issuecomment-2002195393).

But I think that this can be implemented easily on top of your ldap auth and is no case a blocker.

redimp:
Sorry, my time is very limited at the moment 😞 Please don't let this discourage you.

The example works as expected, wonderful!

Will be on vacation in a week, will spent time on this then.

redimp:
I chose to implement the database migrations myself. This works for sqlite, mariadb and postgres.

PR #207, opened 2025-02-20
Frontmatter Plugin / Page Properties!

ReessKennedy:
## Solution
Simple Frontmatter display that toggles & is hidden by default (See more ideas below the image)

## Why
- Frontmatter / page properties supported by most markdown editors in the world
- We had a team repo we wanted to try collaborating on with the basic properties (seen in the below gif) but when we tried to sync it with Otter it wasn't parsed and created a real mess and this made me sad so wanted to fix because Otter is otherwise so great. 

Otter FrontMatter

## More ideas
- Could add something to selectively remove completely from a page if something like `properties: false` is set at the page level AND could (probably should) made this a global setting inside editing settings like, "Hide/show page properties" ..... this would not impact people who don't even know what frontmatter is and wouldn't change appearance of anyone's existing site. 
- Inspirations on design here from the way Obsidian does it in a little toggle ... basically the most popular markdown tool in the world right now, I think, so they are probably make good design decisions
- Piggybacked on existing toggle js used to hide/show content in another plugin and modified the css slightly
- CSS was stuffed in custom but should, of course, be worked into core if someone agrees this should be a core feature

### Larger Screenshot
But see the above gif for behavior on click ... 

Otter Wiki Frontmatter

### DIFF
--- a/otterwiki/renderer_plugins.py
+++ b/otterwiki/renderer_plugins.py
@@ -2,6 +2,7 @@
 # vim: set et ts=8 sts=4 sw=4 ai:
 
 import re
+import yaml
 
 from mistune.inline_parser import LINK_LABEL
 from mistune.util import unikey, ESCAPE_TEXT
@@ -81,7 +82,6 @@ class mistunePluginFootnotes:
     def parse_footnote_item(self, block, k, refs, state):
         def_footnotes = state['def_footnotes']
         text = def_footnotes[k]
-        idx = list(def_footnotes.keys()).index(k) + 1
         stripped_text = text.strip()
         if '\n' not in stripped_text:
             children = [{'type': 'paragraph', 'text': stripped_text}]
@@ -101,7 +101,7 @@ class mistunePluginFootnotes:
         return {
             'type': 'footnote_item',
             'children': children,
-            'params': (k, idx, refs),
+            'params': (k, refs),
         }
 
     def md_footnotes_hook(self, md, result, state):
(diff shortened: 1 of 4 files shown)


redimp:
Hey @ReessKennedy, I tried out your changes, but it looks like something is missing, I still see a line under the properities:

image

Or is my markdown wrong?

```markdown
---
draft: false
date: 2025-02-24
---
# Frontmatter

Testing the frontmatter

```

Side note: Feel free to integrate the css code from the custom css in the `otterwiki/static/css/otterwiki.css`.

ReessKennedy:
Hey @redimp ... you should see it fine now as I moved the CSS into "otterwiki.css" file. 

I also added an additional function that will look if "title" is written into the front matter and render that as the h1 element for the page. 

Neither of these changes will adversely impact anyone's experience since only trigger for those who, like me, but could make a bunch of people happy who already use page meta data like this or want to.

redimp:
Hey @ReessKennedy, thanks for the effort! Will review and test this and I'm pretty sure its ready to merge.

otterwiki/renderer_plugins.py
```diff
@@ -81,7 +82,6 @@ def parse_def_footnote(self, block, m, state):
     def parse_footnote_item(self, block, k, refs, state):
         def_footnotes = state['def_footnotes']
         text = def_footnotes[k]
-        idx = list(def_footnotes.keys()).index(k) + 1
```
redimp, line 84: Please fix: This reverted f8979cba33cb7dfbbfdde0fa422ce5d60b11cc01 by accident.

redimp: I did this locally via `git cherry-pick -n f8979c`, but I'm not allowed to push into this branch.

redimp:
The CSS is still not correct ... the front matter is rendered okay:
image

But the foldable Menus are broken:
image

redimp:
Im went ahead and cherry-picked the commits:

- 0b010dda3c34881f98b63dd31a3db6729e62d150
- e9cdc5a44df3339bd662eccc2c0179fccd657dc2
- 81be1ffdae5a51f1bb59ef42f3493ee69219f03a
- 7dcc7171e8c8ad0225fffd37d45b958af8fd8501

Fixed the css and added a test.

redimp:
Released in [v2.10.5](https://github.com/redimp/otterwiki/releases/tag/v2.10.5).

PR #282, opened 2025-08-19
Author plugin shows creator and last modified metadata

tarxf:
Changes to include the author of the first commit of the file as 'creator' in the authorsignature plugin.

This is a simplistic approach that:

1. locates and appends the creator info in the metadata call from gitstorage.py. This is a choice that would be beneficial if the plan is to eventually promote the author footer into the page view. See alternate proposal below.
2. will miss the original author if the file is moved/renamed. It will show instead the author of the page move. This can be fixed but will require a bit more code and deciding if that's what should be done.

An **alternate proposal**, if the changes to gitstorage.py are invasive, is to instance a git repo object in the plugin and get the data there. If the author footer in the page is not planned to be integrated to the page natively, then this approach might be preferred.

Let me know your thoughts.

**Testing**

Screenshot From 2025-08-19 11-24-46

### DIFF
--- a/docs/plugin_examples/README.md
+++ b/docs/plugin_examples/README.md
@@ -25,6 +25,9 @@ the volume that keeps the `app-data`, where also the `db.sqlite` and the
 `repostitory` lives. Or you organize the plugins in a separate directory
 which you mount in the container into `/plugins`.
 
+In environments with `SELINUX=enforcing` the bind mounts have to be adjusted,
+please check the [FAQ](https://otterwiki.com/FAQ#environments-with-selinux).
+
 *Note:* For plugins with complex dependencies the docker image might be missing
 packages. In this case adding these packages to the image is inevitable, a
 custom image (which packs the plugin, too) might be the best solution.
@@ -32,4 +35,3 @@ custom image (which packs the plugin, too) might be the best solution.
 ### Demo of the example plugins
 
 You can find a `docker-compose.yaml` in [docs/plugin_examples](https://github.com/redimp/otterwiki/tree/main/docs/plugin_examples).
-
--- a/docs/plugin_examples/plugin_authorsignature/otterwiki_authorsignature.py
+++ b/docs/plugin_examples/plugin_authorsignature/otterwiki_authorsignature.py
@@ -13,9 +13,14 @@ from otterwiki.plugins import hookimpl
 @hookimpl
 def page_view_htmlcontent_postprocess(html, page, storage):
     if page.metadata is not None:
+        # Getting the first sha of the page file to request the metadata to be used
+        # for the page creation info.
+        first_sha = storage.repo.git.rev_list("--max-parents=0", "HEAD", page.filename)
+        creation_metadata = storage.metadata(page.filename, first_sha)
         html += f"""
         <div style="margin-top: 5rem; padding-top: .5rem; border-top: 1px dashed rgba(128,128,128,0.2); color: rgba(128,128,128,0.5);" class="text-small">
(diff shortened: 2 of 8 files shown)


otterwiki/gitstorage.py
redimp, line 1: I would like to suggest reconsidering the automatic retrieval of `creator_name` and `creation_datetime` in each call to `storage.metadata()`.

Especially building a list of all commits to just fetch just the author and date of the first one will get expensive for files with a long history.

This first thing that jumps to my mind is to add a implemtentation of `git rev-list --max-parents=0 HEAD filename` using gitpython to the plugin which fetches the hash of the first commit, which can be used via the `metadata()` to retrieve `creator_name` and `creation_datetime`.

What do you think?

redimp: Oh, you are right .. the plugin has no access to the storage object.

Might be a good opportunity to extend either the plugin hook, or add a new that gets more context. Will think about this.

tarxf: moving to WIP 

> Oh, you are right .. the plugin has no access to the storage object.
> 
> Might be a good opportunity to extend either the plugin hook, or add a new that gets more context. Will think about this.

Hi, consider the last push I did to the branch as a draft for your proposal for the hook being extended to make the storage object available.

With this, the plugin contains the logic to gather the creation metadata to show it in the footer. There are no changes to the internals of the otterwiki, other than the addition of the storage object in the plugin signature.

Also, the rev_parse is now being used instead of the iter_commits to get the first commit of the file.

**Testing**
User update.
Screenshot From 2025-08-20 20-55-00

Anonymous update.
Screenshot From 2025-08-20 21-23-03

There is still the option of doing it in a separate plugin but I'll leave that to you since it's an architectural decision.

redimp:
Looks good to me, will do some testing and come back to you!

redimp:
This pull request made me think about the plugin structure currently used in An Otter Wiki:
It's very basic and was mostly made as a proof of concept, with a bit of hope to motivate developers to start creating plugins and requesting features. Well that worked.

This PR reminded me, that currently the plugins have minimal access to anything in the application, as seen here: the `page_view_htmlcontent_postprocess()` hook has no access to e.g. the `storage`.

So I've added the `@hookspec setup()` with the parameters: app, db, storage. Plugins that need any of these can just implement the hook and store the variables in the class, e.g.:

```python
class AuthorSignature:
    @hookimpl
    def setup(self, app, db, storage) -> None:  # pyright: ignore [reportUnusedVariable]
        self.storage = storage
```

This has another effect: The hookspec for `page_view_htmlcontent_postprocess` does not need to be changed and is backward compatible and has access to `storage`:

```python
    @hookimpl
    def page_view_htmlcontent_postprocess(self, html, page):
        if page.metadata is not None:
            # Getting the first sha of the page file to request the metadata to be used
            # for the page creation info.
            first_sha = self.storage.repo.git.rev_list("--max-parents=0", "HEAD", page.filename)
            creation_metadata = self.storage.metadata(page.filename, first_sha)
            html += f"""
            <div style="margin-top: 5rem; padding-top: .5rem; border-top: 1px dashed rgba(128,128,128,0.2); color: rgba(128,128,128,0.5);" class="text-small">
                Last modified {page.metadata["datetime"].strftime("%Y-%m-%d %H:%M:%S")} by {page.metadata["author_name"]}.
                <br/>Created {creation_metadata['datetime'].strftime("%Y-%m-%d %H:%M:%S")} by {creation_metadata['author_name']}.
            </div>
            """
        return html
```

redimp:
Released in [v2.14.0](https://github.com/redimp/otterwiki/releases/tag/v2.14.0).

Issue #283, opened 2025-08-19
entrypoint.sh plugin installer is not expanding plugin dirs


tarxf:
The code in entrypoint.sh to check if /app-data/plugins or /plugins has directories inside is not finding any dirs inside. This results in no plugin gets installed even if these dirs exist in the designated folder.

Instructions to reproduce:

1. In a fresh clone of this repo,
2. Build the container:
   ```podman build -t plugin-test .```
3. Run the container with docs/plugin_examples shared in the container as /plugins
   ```podman run -v ./docs/plugin_examples/:/plugins -p 8080:8080 plugin-test:latest```
4. Load the Otterwiki in the browser ```http://localhost:8080```

**Expected result**: The author plugin loads at the end of the page. 

**Actual result**: The page loads without the author plugin.

With some debugging, I figured that [the shell expansion in the for loop from entrypoint.sh](https://github.com/redimp/otterwiki/blob/main/docker/entrypoint.sh#L98) is not happening, therefore the code does not proceed to install it. I have a fix [in this branch ](https://github.com/redimp/otterwiki/compare/main...tarxf:otterwiki:feature/portable-entrypoint-plugin-install) that I've tested it works and I'll submit as PR.

redimp:
Hey @tarxf, thanks for reporting this issue. Will do some testing and check your PR.

redimp:
I can not reproduce the error you are reporting.

## podman on rocky9 using otterwiki:2.

```
redimp@rocky9 ~/otterwiki $ podman pull redimp/otterwiki:2.13.0
[...]
redimp@rocky9 ~/otterwiki $ podman run -v ./docs/plugin_examples/:/plugins:Z -p 8080:8080 otterwiki:2.13.0
Initialized empty Git repository in /app-data/repository/.git/
Installing: /plugins/plugin_authorsignature/
Processing /plugins/plugin_authorsignature
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
Building wheels for collected packages: otterwiki_authorsignature
  Building wheel for otterwiki_authorsignature (pyproject.toml): started
  Building wheel for otterwiki_authorsignature (pyproject.toml): finished with status 'done'
[...]

PR #284, opened 2025-08-19
entrypoint.sh: make plugin install more portable

tarxf:
Addresses #283 

Making the plugin discovery more predictable by adding a second layer that calls the individual plugin directories in the predefined locations.


redimp:
With the underlying issues in #283 being identified this PR is not necessary. Thank you for the effort nonetheless.

PR #298, opened 2025-10-16
feat: custom stylesheet management

hollisticated-horse:
Introduce a guided Custom.css workflow: load the stylesheet from the wiki, seed new installs with a commented template, hide it from the page index, and surface quick links in user settings. Restrict access so only admins or writers can edit or view the raw page while the rest of the site consumes the compiled CSS. The editor now treats Custom.css as real CSS with syntax highlighting and a streamlined UI.

Full disclosure, this PR was written via ChatGPT Codex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- insertion marker -->
+## [v2.14.2](https://github.com/redimp/otterwiki/releases/tag/v2.14.2) - 2025-11-09
+
+<small>[Compare with v2.14.1](https://github.com/redimp/otterwiki/compare/v2.14.1...v2.14.2)</small>
+
+### Features
+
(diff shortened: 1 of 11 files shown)


redimp:
Hey @hollisticated-horse, thanks for the PR!

Really like this idea. I'll need some time to review it fully, so please be patient.

Just one quick question: Why is the 'Custom.css' file created by default in `server.py`? Would it be better to create it only when saving it, after opening it from the settings? It should just be saved like other pages? Let me know your thoughts!

hollisticated-horse:
> Really like this idea. I'll need some time to review it fully, so please be patient.

Cheers, no worries about time. 

> Just one quick question: Why is the 'Custom.css' file created by default in `server.py`? Would it be better to create it only when saving it, after opening it from the settings? It should just be saved like other pages? Let me know your thoughts!

Good question. My thinking was to create it by default with relevant selectors already in the file, for quicker onboarding. But if we can do that when an admin/user creates the custom.css file from within the UI, I guess that works too. 

EDIT: I had forgotten: I made an initial custom.css file, with the selector commented out. This file is copied to custom.css when custom css is activated. Easier restore file. (But I didn't implement the reset).

redimp:
Hey @hollisticated-horse, I thought about your PR some more and I'm still hesitant about the introduced complexity for one customization. I would like to generalize your idea.

Please share your thoughts on possible solutions in this discussion https://github.com/redimp/otterwiki/discussions/389.

PR #309, opened 2025-11-17
Fix: stop escaping anything when calling preview

### LINKED ISSUE #307: Broken preview requests for pages with apostrophes
1. Create a page with apostrophe, for example `TestPagewith'`.
2. Try loading the preview. `/TestPagewith&#39;/preview` will be requested and it will fail with a generic 'Method Not Allowed' error.
3. Requesting a preview with `/TestPagewith'/preview` (without encoding) works.

Not sure if it's a browser/frontend problem or my nginx configuration, any ideas?

deseven:
This should fix #307.

My line of thought is the following: there is no point in escaping anything when calling a preview since it works with any page names, including the ones that don't exist and even can't exist, simply for building the breadcrumbs. XSS doesn't seem to be possible here either, because everything is being sanitized in the preview output anyway.

### DIFF
--- a/otterwiki/templates/editor.html
+++ b/otterwiki/templates/editor.html
@@ -588,7 +588,7 @@ thx to https://github.com/sparksuite/simplemde-markdown-editor/issues/328#issuec
         formData.append("cursor_ch", cm_editor.getCursor().ch);
 
         /* FIXME: display loader */
-        fetch("{{ url_for('preview', path=pagepath) }}", {
+        fetch("{{ url_for('preview', path=pagepath) | safe }}", {
                 method: 'POST',
                 body: formData,
             })


redimp:
Hey @deseven, thanks for the PR. 

This fixes the issue with a single quote for the preview, but I have to do some more testing to be sure not to break something else. 

I would prefer to correctly urlencode all characters in hrefs, e.g. a single quote `'` as `%27`. Working on that.

redimp:
Using `{{ url_for(..) | safe }}` solves the Problem for `'` in page paths, but breaks for `"`. 

The proper solution is to use [urlencode](https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.urlencode).

I have a version ready for testing, will continue to work on this tomorrow and hopefully release a new version till the end of the week.

PR #314, opened 2025-11-24
Sitemap generation

deseven:
This adds sitemap generation at `/sitemap.xml` that could be added to Google Search Console (or other search engines) in order to help with the indexing. Basically it builds everything from existing markdown files, adding priorities based on the depth level.

### DIFF
--- a/otterwiki/sitemap.py
+++ b/otterwiki/sitemap.py
@@ -34,6 +34,10 @@ def sitemap():
             else:
                 page_url = url_for('view', path=pagepath, _external=True)
 
+            # post-processing to encode the quotes
+            page_url = page_url.replace("'", "%27")
+            page_url = page_url.replace('"', "%22")
+
             loc = SubElement(url_elem, 'loc')
             loc.text = page_url
 
@@ -43,13 +47,13 @@ def sitemap():
 
             # Calculate priority based on depth
             # 1.0 for root, 0.9 for level 1, 0.8 for level 2, etc.
-            # Minimum is 0.5
+            # Minimum is 0.1
             priority = SubElement(url_elem, 'priority')
             if pagepath.lower() == 'home' or pagepath == '':
                 depth = 0
             else:
                 depth = pagepath.count('/')
-            calculated_priority = max(1.0 - (depth * 0.1), 0.5)
+            calculated_priority = max(1.0 - (depth * 0.1), 0.1)
             priority.text = f'{calculated_priority:.1f}'
 
         except Exception as e:
@@ -57,7 +61,9 @@ def sitemap():
             continue
 
     indent(urlset, space="  ", level=0)
-    xml_string = tostring(urlset, encoding='utf-8', xml_declaration=True)
+    xml_string = tostring(
+        urlset, encoding='utf-8', method='xml', xml_declaration=True
+    )
     response = make_response(xml_string)
     response.headers['Content-Type'] = 'application/xml; charset=utf-8'
     return response
--- a/tests/test_sitemap.py
+++ b/tests/test_sitemap.py
@@ -9,11 +9,5 @@ def test_sitemap(admin_client):
     response = admin_client.get("/sitemap.xml")
     assert response.status_code == 200
     assert response.headers['Content-Type'] == 'application/xml; charset=utf-8'
-    assert (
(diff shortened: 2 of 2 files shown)


redimp:
Hey @deseven, thanks for this PR!

Can I ask you for a chore, a bit of refactoring: Please move the implementaton into a new file `otterwiki/sitemap.py` that is imported in `views.py`?

deseven:
@redimp sure. I've also added a test, would you like it to be moved out of essentials as well?

Also, do we aim for a minimum python version? There's a fallback for prettifying on python < 3.9, but not sure if it's actually needed.

redimp:
Not that important, but for the sake of completeness: Yes please.

Thanks for all the effort!

redimp:
Dont worry about python 3.9, we require

https://github.com/redimp/otterwiki/blob/2ada269960cebac1489525a537a5219763715ecd/pyproject.toml#L9

deseven:
Oh well, I've checked for Python version in the readme and in the wiki, but didn't check in the most obvious place :)

All done.

redimp:
Looks good from my end, though I’ll admit my SEO knowledge isn’t extensive. I did a quick scan of the [Sitemaps Protocol](https://www.sitemaps.org/protocol.html), and under “Entity escaping,” it specifies that URLs must be properly escaped. Let’s stick to the standard and encode both page names and URLs accordingly.

Also, why limit the priority to a minimum of 0.5? The official docs say priority values can range from 0.0 to 1.0 — so we might want to allow lower values if needed.

Thanks for working on this!

redimp:
Added the branch to my test image, see <https://testing.otterwiki.com/sitemap.xml>

deseven:
> I did a quick scan of the [Sitemaps Protocol](https://www.sitemaps.org/protocol.html), and under “Entity escaping,” it specifies that URLs must be properly escaped. Let’s stick to the standard and encode both page names and URLs accordingly.

Well, this is tricky. XML specs do not require single quote escaping and everything else except for the ampersand is being escaped differently by Flask's `url_for()`. This should be enough already since we produce valid XML with valid escaped URLs, but to match the provided substitution table exactly I guess the only way would be to generate the XML manually, including the URLs?

In general I'm not sure how to approach this and the XML examples they provide are incomplete since they only show how to escape the ampersand which is already being done. I'll try to look into other `sitemap.xml` generators to see how they handle it, but quick googling suggests that it's open to interpretation.

> Also, why limit the priority to a minimum of 0.5? The official docs say priority values can range from 0.0 to 1.0 — so we might want to allow lower values if needed.

No particular reason, really. The priority is a suggestion and the implementation (if there's any at all) depends on a search engine. I'll make the lower limit 0.1 just in case some search engines simply discard the pages with 0 priority.

deseven:
As an example:
```python
from xml.etree import cElementTree as ET
text = "Test string with < > ' \" &"
root = ET.Element('test')
root.text = text
ET.tostring(root, method="xml")
```
`<test>Test string with &lt; &gt; ' " &amp;</test>`

deseven:
Okay, after digging into this further it only got more muddy :)

The sitemaps standard mentions that the URLs must follow RFC 3986, but html escape codes such as `&amp;` are not compatible with it. I think the confusion comes from mixing the URL format with the XML data format itself. It doesn't look like it makes sense to write our own XML implementation to follow the escaping exactly and I don't see other implementations doing so. In general nobody seems to care much, standard libraries are being used and that's it.

But if you look at it from another angle, Flask's `url_for()` already encodes the URLs into percent-encoding and this is what the crawler would see on the website itself. In order to follow the sitemaps standard exactly we could simply manually encode single and double quotes and the ampersand will be handled automatically by the XML library according to the XML format specs.

So for example, the page called `Page With ' " < > & For Testing` will be
 - encoded to `Page%20With%20'%20%"20%3C%20%3E%20&%20For%20Testing` by Flask
 - encoded to `Page%20With%20%27%20%22%20%3C%20%3E%20&%20For%20Testing` by us to follow RFC 3986
 - saved to xml as `Page%20With%20%27%20%22%20%3C%20%3E%20&amp;%20For%20Testing`

As far as I can tell from using various sitemap validators this is correct:
image

What do you think?

redimp:
Looks good to me!

Thanks for digging into this and double checking.

redimp:

PR #315, opened 2025-11-28
Thumbnail size selection

deseven:
I've noticed that the backend already supports defining max thumbnail size by passing the value to `thumbnail` param. This PR adds this functionality to the frontend by providing an optional size field:
image

Additionally, it fixes a small bug when selecting different image attachments one by one leads to multiple image previews. Not sure if it's intended, tell me if it is and I'll remove it.


redimp:
Looks good to me. Will check and merge later. Also planning to make a new release later today.

redimp:

PR #316, opened 2025-11-28
Setting to treat underscores as spaces for page titles

deseven:
Not sure if you'd like this idea, but please tell me what you think.

Right now otterwiki has no distinction between page name and page title, but this could be useful if you want to have clean URLs and page names with spaces. So I've added a setting to treat underscores as spaces in various menus. This would allow us to have urls such as `/My_Page`, while seeing human-readable `My Page` in the menu. This is basically what MediaWiki does.

### DIFF
--- a/otterwiki/helper.py
+++ b/otterwiki/helper.py
@@ -206,6 +206,23 @@ def get_pagename(filepath, full=False, header=None):
     return "/".join(arr)
 
 
+def get_pagename_for_title(filepath, full=False, header=None):
+    '''this will derive the page name for display purposes (titles, breadcrumbs, etc.) with underscore replacement'''
+    pagename = get_pagename(filepath, full=full, header=header)
+
+    if app.config.get("TREAT_UNDERSCORE_AS_SPACE_FOR_TITLES", False):
+        if full:
+            # replace underscores in each path component
+            parts = pagename.split("/")
+            parts = [part.replace("_", " ") for part in parts]
+            return "/".join(parts)
+        else:
+            # replace underscores in the single component
+            return pagename.replace("_", " ")
+
+    return pagename
+
+
 def get_pagename_prefixes(filter=[]):
     pagename_prefixes = []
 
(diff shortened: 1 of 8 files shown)


redimp:
You are on fire! Will review this tomorrow, unfortunately not enough brain power today, to make a release.

deseven:
image

deseven:
> You are on fire! Will review this tomorrow, unfortunately not enough brain power today, to make a release.

Sorry, I'm just in the all-in mode for the wiki project I'm doing. No pressure :)

redimp:
From my observation many people dislike the `%20` in browser urls and favor the `_` in between words. I thought a lot about adding some magic, where a `_` in the url is converted into a ` ` (space) on the fly, but could not decide how to handle file names.

Your approach is less magic and with the settings users can decide how they want to handle their urls.

Hm, but hinking about it, would it be convenient if a user in An Otter Wiki with `TREAT_UNDERSCORE_AS_SPACE_FOR_TITLES=1` adds a page "New York" that the page is created with the path `New_York` and the title `# New York`?

redimp:

deseven:
> Your approach is less magic and with the settings users can decide how they want to handle their urls.

I was thinking about different approaches too, my original idea was to grab page title from the page markdown itself (first `#` header), which would've been extremely flexible for all possible use-cases, but this turned out to be [a huge effort that required touching a lot of core wiki functionality](https://github.com/deseven/otterwiki/commit/81a0dfd1d199663c2f7ad35fdcfcbbc7a474b7b1), I wasn't comfortable with it and it didn't feel like it respects otterwiki vibes :) So yeah, the current solution is a compromise of sorts.

> would it be convenient if a user in An Otter Wiki with TREAT_UNDERSCORE_AS_SPACE_FOR_TITLES=1 adds a page "New York" that the page is created with the path New_York and the title # New York?

Hm, looking at it from a user perspective, I would not expect this setting to lead to this behaviour. I think maybe a separate setting to replace spaces with underscores when creating pages would be better? Because technically these 2 options are not in conflict with each other. If you think this has value I'll make another PR with it.

Issue #317, opened 2025-11-29
Thumbnails link doesnt repsect width


gomaaz:
Hello, 

I guess this feature doesnt work correctly by now.
The pasted link doesnt set a pixel size:
It doesnt matter which value I type in (500px or 500)
The links look always like this 

`[![](./image-1764350608197.png?thumbnail)](./image-1764350608197.png)`

Version 2.14.3 (v2.14.2-0-gd6d5f3c)

<img width="272" height="248" alt="Image" src="https://github.com/user-attachments/assets/e61ee977-7c0b-4a9f-aff8-34392275e9b5" />

https://github.com/redimp/otterwiki/pull/315

redimp:
Hey @gomaaz, thanks for reaching out.

Can not reproduce ... which browser and which version are you using?

Just tested on <https://demo.otterwiki.com/thumbnail/edit>

<img width="322" height="229" alt="Image" src="https://github.com/user-attachments/assets/5b69c355-12ab-4b71-ae29-dd8be73586fb" />

pasted

```
[![](./image-1764418394515.png?thumbnail=25)](./image-1764418394515.png)
```

gomaaz:
Latest version for chrome.
Weird, now it's working, even for my docker-instance...

watched the browser-console for any logs, but nothing showed up.

Anyhow, sorry for the interruption.

Great piece of software! 👌

redimp:
Thank you for double checking!

deseven:
I thought that maybe the old javascript was cached in your browser, but the wiki currently requests `otterwiki.js` with a unique id, so that shouldn't be the case...  I guess it is possible that the html markup was cached as well.

Please report if this happens again on your current install and I'll look into it more.

PR #318, opened 2025-11-29
Small fixes

deseven:
Fixes for my previous PRs (sorry):

1. Do not apply underscore replacement for page rename form. Currently when you try to rename a page called `Test_1/Test_2` with `TREAT_UNDERSCORE_AS_SPACE_FOR_TITLES=True`, current path is being displayed as `Test 1/Test 2`, which is incorrect.

2. For thumbnail size field, it's currently possible to write anything that is not a number (attempted in #317), so as a quick fix it's now only possible to input positive numbers.


redimp:
Hey @deseven, will check and merge later today.

redimp:
Released in [v2.14.4](https://github.com/redimp/otterwiki/releases/tag/v2.14.4).

PR #323, opened 2025-12-12
Optimized resource loading

deseven:
Prevents loading of Mermaid and MathJax unless needed (required to render the page). The logic determines what needs to be loaded when, for editor and syntax guide both are always loaded.

Decided to do everything on the backend to be able to cover it with tests.


redimp:

PR #326, opened 2025-12-12
Custom head and body html from files

deseven:
Adds `customHead.html` and `customBody.html`, in addition to `custom.js`/`custom.css`. There's also a test that checks both html injection methods.


redimp:

PR #327, opened 2025-12-14
Fix for broken revision links after page rename

### LINKED ISSUE #306: Broken history after page rename
Noticed an issue with page history - if the page was renamed at some point of its life, the links to specific revisions become broken.

Consider the following: https://strixhalo.wiki/Hardware/PCs/GMKtec-EVO-X2/history

The page was renamed 2025-06-24 17:35:08 in the commit `8f88aa` and none of the links to revisions earlier than that work (for example [35145e](https://strixhalo.wiki/Hardware/PCs/GMKtec-EVO-X2?revision=35145e), they return 404 instead and this **might** hurt the indexation severely, as Google for example marks the website as containing too many errors and refuses to index it.

Ideally these links need to work, but alternatively at least don't return 404 for them and instead output the latest revision of the page with some note that this revision currently cannot be accessed.

deseven:
My attempt to fix #306

The idea is to generate correct page path that the page had before the rename - git will handle the rest and return correct content for that page at the specified revision.

The included tests cover single and multiple renames.

### DIFF
--- a/otterwiki/wiki.py
+++ b/otterwiki/wiki.py
@@ -380,14 +380,39 @@ class Page:
     def load(self, revision: str | None):
         metadata = None
         content = None
-        try:
-            content = storage.load(self.filename, revision=revision)
-            metadata = storage.metadata(self.filename, revision=revision)
-        except StorageNotFound as e:
-            if all([not metadata, not content]):
-                # If both are None, raise the exception. Otherwise, a warning will show on the page that
-                # the file is not under version control.
-                raise e
+
+        # if a revision is specified, we need to handle potential previous file renames
+        if revision is not None:
+            try:
+                # first try with the current filename
+                content = storage.load(self.filename, revision=revision)
+                metadata = storage.metadata(self.filename, revision=revision)
+            except StorageNotFound as original_error:
+                # if that fails, try to get the filename that was used at this revision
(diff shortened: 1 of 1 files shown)


deseven:
Oh, I found a problem with this approach: while the revision page does work, all other functionality does not (blame, history, rename, delete), because there technically no such page and it's correct.

Ideas on how to solve this:
 - disable all that functionality for revision page
 - instead of replacing the full path, add an argument that would set the old name (URL would look like `/current/path?revision=d13dc9&oldpath=%2Fold%2Fpath`

deseven:
Gonna work on it more.

deseven:
Okay, how about a different, simpler approach? Instead of changing the page path we'll handle everything internally. This won't show the old path to the user, but it will show the correct page state at revision no matter if it was renamed or not. I think this is a reasonable compromise.

redimp:
> I think this is a reasonable compromise.

It's a great approach if the goal is a flawless user experience.

deseven:
Alright, then all good (implemented in my last commit).

redimp:
Well engineered! Thank you for all these awesome contributions.

redimp:

deseven:
Thank you for having such a cool project :)

Issue #329, opened 2025-12-22
Custom HTML files cannot be loaded inside the container


deseven:
I've noticed that in the docker image the app is actually being installed in `/opt/venv` instead of `/app`, so [trying to find custom html files in `app.root_path`](https://github.com/redimp/otterwiki/blob/main/otterwiki/helper.py#L369) is incorrect. Current implementation doesn't work with what is [written in the documentation](https://otterwiki.com/Customization) when wiki is being run in a container.

I see the variable `USE_STATIC_PATH` in `entrypoint.sh` which could be useful in that case, but I'm unsure what would be a clean way to pass it in supervisord/uwsgi. I tried to simply export it, but it doesn't seem to propagate to uwsgi. The obvious way would be to add `env=USE_STATIC_PATH=/app/otterwiki/static` to `uwsgi.ini`, but hardcoding this path might not be ideal since `entrypoint.sh` allows overriding it with `STATIC_PATH`.

@redimp your thoughts?

deseven:
As a workaround for now (if someone runs into the issue), you can mount the files directly using the correct path, for example in docker-compose:
```yml
    volumes:
      - /some/path:/app-data
      - /some/path/custom:/app/otterwiki/static/custom
      - /some/path/custom/customHead.html:/opt/venv/lib/python3.11/site-packages/otterwiki/static/custom/customHead.html
      - /some/path/custom/customBody.html:/opt/venv/lib/python3.11/site-packages/otterwiki/static/custom/customBody.html
```

redimp:
Fix released in [2.15.1](https://github.com/redimp/otterwiki/releases/tag/v2.15.1).

deseven:
Tested all 3 scenarios: running on host, running in container, running in container with slim image. All seems good 👍

PR #331, opened 2025-12-23
load_custom_html respects USE_STATIC_PATH

redimp:
this solves the issues with load custom body and header files, see #329

My first idea how to solve the issue with loading custom body and header files from the correct directory, that came up in #329:

Make the entrypoint export the `USE_STATIC_PATH` variable and use it in `load_custom_html()` when loading `customHead.html` and `customBody.html`.

### DIFF
--- a/docs/custom_html_example/README.md
+++ b/docs/custom_html_example/README.md
@@ -1,0 +1,6 @@
+# Customizing (tweaking) an Otter Wiki
+
+This is an example for tweaking the HTML of An Otter Wiki.
+
+Please check <https://otterwiki.com/Customization> for
+more examples.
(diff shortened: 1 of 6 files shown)


redimp:
What do you think about this approach @deseven?

(Please note, due to the holidays and family my availability will be very limited in the next days.)

deseven:
Hm, this is exactly [what I tried myself](https://github.com/deseven/otterwiki/commit/b2459e4885cf3191b7a1c662461a945d6f45c46a), but this didn't work because this export didn't appear in the uwsgi's env.

I guess I messed up when testing it, let me try again.

Happy holidays :)

deseven:
Yes, this totally works!

Not sure what happened, maybe I forgot to rebuild the image previously.

redimp:
Thanks for double checking. Still on my TODO list is to fix this for the -slim image.

PR #336, opened 2026-01-04
More plugin hooks for html injection & processing

deseven:
Currently you only have a way to grab the whole rendered page in html, which could be problematic if you need to alter specific elements (in that case you end up with either using regexes or parsing the whole html to do what you want - not ideal).

There is also no way to inject additional resources or scripts into the page, which I believe a plugin should be able to do.

So this PR adds 5 additional hooks:
1. `template_html_head_inject` - Inject content into `<head>`
2. `template_html_body_inject` - Inject content into `<body>`
3. `renderer_process_link` - Process links during markdown rendering
4. `renderer_process_image` - Process images during markdown rendering
5. `renderer_process_heading` - Process headings during markdown rendering

Obviously more could be added, [the list is quite large](https://mistune.lepture.com/en/latest/renderers.html#available-methods), but I think this would do for now.

Any thoughts?

### DIFF
--- a/docs/plugin_examples/plugin_htmlinjection/otterwiki_htmlinjection.py
+++ b/docs/plugin_examples/plugin_htmlinjection/otterwiki_htmlinjection.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+
+"""
+HTML Injection Example Plugin for An Otter Wiki
+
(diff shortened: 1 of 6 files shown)


redimp:
This looks like a good step in the direction of having more (useful) plugin hooks. Will do some testing and merge this asap.

PR #338, opened 2026-01-06
Automatic plugin uninstallation

deseven:
This removes all plugins that were removed from `/plugins` by the user, creating a better user experience (I hope), where you can just add or remove needed plugins without thinking to much of internal stuff.

I didn't touch slim image yet, in case you have other ideas on how to approach it.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -94,13 +94,43 @@ echo "    location / {
 # close the server block
 echo "}" >> /etc/nginx/sites-enabled/default
 
+INSTALLED_PLUGINS=$(pip list --format=freeze | grep -E '^otterwiki[_-]' | cut -d'=' -f1 || echo "")
+EXPECTED_PLUGINS=""
+
 # install plugins found in /app-data/plugins and /plugins
 for PLUGIN in /app-data/plugins/*/ /plugins/*/; do
     test -d "$PLUGIN" || continue
     echo Installing: $PLUGIN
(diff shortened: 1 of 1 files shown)


redimp:
While runtime package installation isn't ideal, it offers a valuable convenience by enabling users to run or test plugins without requiring them to build their own images. 

However, I don't recommend attempting to uninstall packages during startup. Runtime installations can leave behind residual files and dependencies, potentially leading to unexpected issues. For a clean and predictable environment, users should recreate the container from the base image when a plugin needs to be removed or the runtime environment rebuilt. This ensures a fresh start and avoids potential conflicts caused by leftover package remnants.

deseven:
Alright, scrapping it then.

Still, right now the situation could be quite confusing for the end user - you put a plugin in the specified directory, the plugin automagically starts working, then you remove it and yet it continues working. I'd at least expect it to be described in the docs, that the user is expected to recreate the container.

redimp:
I agree, this is behavior is confusing. I updated the documentation and hope this helps to make it more clear.

deseven:
If I may add :)

> When running An Otter Wiki from a virtual environment run `pip uninstall <plugin-name>`. When running from a docker image it is recommended to recreate the container for a clean environment **after removing the plugin from the plugins dir**.

Or something like that...

redimp:
Done in https://github.com/redimp/otterwiki/commit/059e9c59f5092be9a3dcb0d78589661c140142ea .

Issue #24, opened 2022-10-05
Push wiki to remote


marcrleonard:
Is there a place in the interface the push the wiki to a git remote location? It would be cool if there was a "Push" or "Sync" feature. This could be very tricky if there are remote changes that conflict with local, but it would great to do this in the UI

redimp:
The major problem that I see here is authentication. When pulling from a remote this might be via https and possible as an anonymous user, but when it comes to pushing you need to be authenticated in most of the cases. To implement this you would have to take care of a huge pile of problems.

marcrleonard:
Yeah I was kind of thinking there would be an initial requirement that their credentials would have to be cached with the os. Then, presumably, the git api would be able to push. 

If a user would try to push (via a button in the ui) and it fails, you could display a helpful error and a 'set up' page outlining how to get the repo/creds set up. You could make it like how an empty GitHub repo will show you the initial set of commands to run to get up and going.

ezcafe:
Please add this feature! I want to backup wiki to Github account

ethereal-engineer:
I need this feature to exist. It would be easiest to achieve using public-key authentication with the git remote. But no, it's not a trivial feature. Damn.

SnowyJaguar1034:
I too would appreciate it if such a feature existed.

ReessKennedy:
+1

Wrykoulaka:
+1, just tried to achieve it mapping ssh keys and adding a post-commit hook but it didnt work. I will setup a cronjob for the time being and call it a day

tarxf:
> +1, just tried to achieve it mapping ssh keys and adding a post-commit hook but it didnt work. I will setup a cronjob for the time being and call it a day

Hey, did your cron job work? Can you post the solution?

I made the git hook work to push on save but I'm having second thoughts about the approach. The save action gets extra latency, plus the repo push is now in the Save success path. In other words, if your git provider is having issues, then you'll see an error on save.

I set it up, it's working fine, but I'll look for an asynchronous approach later. That said, it doesn't need to radically change from the way I designed it.

Let's get into how I did it, I use the **uwsgi installation** and the way I solved it is:

1. Create a repo in a git provider (I use codeberg).
2. Add that remote as an origin to the wiki repo.
3. Setup the ssh keys that the instance will use to push. This includes adding the public key in the git provider.
4. Install the post-commit hook, I am using [this one](https://xn--izc.com/posts/git-autopush/).
[...]

PR #340, opened 2026-01-07
Repository management

deseven:
This should solve #24.

Changes:
1. New admin page "Repository Management".
2. "Enable Git Web server" moved to that new page, I think it makes sense to have it there.
3. New "Enable automatic pushing to SSH remote" setting with a form that allows to define remote URL and optionally a private key. For security reasons these settings cannot be set via env variables. No additional repository configuration is required.
4. Automatic pushes are being triggered on any wiki edit that triggers a commit.
5. `openssh-client` added to both images.

I was also considering to support local remotes and http ones, as well as pulls, but decided to keep things as simple as possible for now.

Tested with regular/slim images and natively (macOS), but I'd appreciate if someone else with clean head could test it thoroughly too :)

### DIFF
--- a/otterwiki/repomgmt.py
+++ b/otterwiki/repomgmt.py
@@ -4,7 +4,7 @@
 import os
 import tempfile
 import stat
-from threading import Thread
+from threading import Thread, Lock
 
 
 class RepositoryManager:
@@ -12,6 +12,8 @@ class RepositoryManager:
     Repository management functionality for Git operations including
     push/pull operations, SSH key management, and periodic pull scheduling.
     """
+
+    git_push_pull_Lock = Lock()
 
     def __init__(self, storage):
         """Initialize with a GitStorage instance."""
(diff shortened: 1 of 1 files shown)


deseven:
Testing process below.

To setup:
```bash
git clone git@github.com:deseven/otterwiki.git
cd otterwiki
git fetch origin push_to_remote
git checkout push_to_remote

docker build -f Dockerfile -t otterwiki-dev .
# or 
#docker build -f docker/Dockerfile.slim -t otterwiki-dev .

docker run \
  --name otterwiki-dev \
  -p 8080:8080 \
  -v ./app-data:/app-data \
  otterwiki-dev

# or
#make debug
```

Generate a new key for authentication:
```bash
ssh-keygen -t ed25519 -f ~/.ssh/id_otterwiki_dev
```

Grab public and private parts:
```bash
cat ~/.ssh/id_otterwiki_dev
cat ~/.ssh/id_otterwiki_dev.pub
```

Go to GitHub or elsewhere and create an empty repo. Grab the remote for it, for example `git@github.com:username/repo.git`.

Then go to the repo settings > deploy keys > add deploy key and paste the public part. Don't forget to check "Allow write access".
image

As the last step go to otterwiki settings, Repository Management, check "Enable automatic pushing to SSH remote" and fill both fields with your remote URL and private key.

After that just try editing the wiki normally. It shouldn't matter if your wiki repo had any commits previously or not. Monitor wiki logs for any problems.

redimp:
Nice! Will do some testing on the weekend. ✌️

redimp:
I've only tested the `ssh://..` connection method so far, and it's working as expected. 

However, the current synchronous push operation seems to slow down editing.  It might be beneficial to offload this process to a background thread, similar to how the threaded mail sending is implemented in `send_mail()` (see: https://github.com/redimp/otterwiki/blob/62b039eaefd2763735a661ace8218bc35a13d430/otterwiki/helper.py#L102).

Furthermore, users may prefer manual synchronization in some situations. Providing a manual sync option through the `Repository Management` interface (e.g., a button) could improve user experience.

deseven:
Should I also implement pulls then? And force pushes?

redimp:
Great. 

Let's require explicit confirmation from the user before performing a force push. 

With pulls we user could run into merge conflicts, and resolving those conflicts is a complex challenge. What's your approach to handling potential merge conflicts that might arise?

deseven:
Merge conflicts (or basically any failures to sync) should be sent to the admin's email since they are for sure important enough. Maybe with a timeout, like no more than one error in an hour. Solving them automatically or providing an interface to do so would be overkill, I think.

Pulling could be executed periodically, every X minutes. And I'm also thinking about providing a unique URL for a pull webhook - calling this URL would trigger a pull, as a signal that there were changes to the repo.

From the UX side I have this on my mind:
```
[ ] Enable automatic pushing to SSH remote
    Remote URL
    Private key
    [Push] [Force Push]

[ ] Enable automatic pulling from SSH remote
    Remote URL
    Private key
    Pull interval (in minutes, 0 to disable)
    Pull webhook URL [copy]
    [Pull]
```
That way you would even be able to create chains if you wish to - pulling from one repo, pushing into another.

**UPD:** Or, actually, the action buttons could appear (or activate) below the form, depending on what user set up.

Any comments?

redimp:
> Merge conflicts (or basically any failures to sync) should be sent to the admin's email [..]

Yes, make sure that the admins don't get spammed. It would also be nice to see success/error in the admin interface. I have to think about how we should store this.

> Pulling could be executed periodically, every X minutes. And I'm also thinking about providing a unique URL for a pull webhook - calling this URL would trigger a pull, as a signal that there were changes to the repo.

Thats a very good idea. I'm thinking about adding an api for ages, when you add this please use the the route `/-/api/v1/...`

> From the UX side I have this on my mind:

The draft looks good to me!

deseven:
Going to reopen when ready.

deseven:
@redimp 

After evaluating this further:

1. Implementing automatic pull interval adds complexity to the wiki while holding little value - pulling on update via webhook is the _proper way_ anyway and nothing would stop you from putting `curl https://example.com/-/api/v1/pull/123456abcdef` in your crontab to achieve periodic pulls if you wish to solve it like that. I suggest dropping automatic pull every X minutes.

2. Same goes to notification emails - if pushes and pulls are solved in a proper way, there won't be an email every minute and there won't be user confusion as to why the error X triggered the email notification, but error Y didn't. In general if we're going to add notification throttling I'd solve it on a higher level for *all* notifications with some sort of a setting like "Do not send more than one email per X minutes" or something.

deseven:
I think this is good to go at this point. I obviously couldn't test every possible scenario, but with a single repo on github it worked out nicely and I was able to edit the content from both sides with very little delay.

redimp:
Not done with the code review yet, but ran a bunch of tests and did not see anything break. With the background push I noticed no delays, 

My only concern is currently that concurrent updates (from multiple users) could cause problems, but thats what the "experimental" label is for.

I hope I manage to review this and make a release till the end of the week.

deseven:
> My only concern is currently that concurrent updates (from multiple users) could cause problems

I've been thinking about it as well, it seems that git does implement several locking mechanisms out of the box to protect the repo from concurrent operations. As far as I understand, running into race conditions here is unlikely and the tree should be safe in any case.

If that assumption is incorrect and someone will run into problems later, I suppose some sort of a git operations queue could be organized, with user-interactable stuff pushed to the top of the queue.

redimp:
I added locking to the push_from_remote() and pull_from_remote() functions. All my tests were good.

redimp:
Github does not let me merge this .. will merge manually.

redimp:
Released in [v2.16.0](https://github.com/redimp/otterwiki/releases/tag/v2.16.0).

PR #349, opened 2026-01-14
Add path attribute to pages and indexes

freddyheppell:
I wanted to hide the table of contents on the homepage and a couple of other pages, but rather than add some complexity for just that, I thought it might be better to add something to allow targeting any specific page with CSS.

This is a fairly common thing in wiki and CMS software e.g. MediaWiki adds `.page-PAGE_NAME`, WordPress adds `.page-ID`.

Because Otter Wiki allows spaces in page names, it's a bit difficult to make valid class names. It could replace spaces with underscores, but I didn't like that approach as it perhaps implies MediaWiki-like behaviour. Instead I've gone for a data attribute, which can contain spaces just fine.

This PR sets `data-page-path`/`data-index-path` set to the `pagepath` template variable, which will give values like:

* `data-page-path="page"`
* `data-page-path="some cat/page"`
* `data-index-path="some cat"`
* `data-index-path="/"` for the global page index

These can then be targeted with CSS, such as:

```css
/* Target specific page */
body[data-page-path="some category/page"] {}

/* Target any page in a category */
body[data-page-path^="some category"] {}

/* Hide homepage toc */
body[data-page-path="home"] .extra-nav {
    display: none;
}
```

The values are always lowercased since accessing a page with any capitilisation (even with "Retain page name case" enabled, at least on *nix?) resolves to the page, but `pagepath` keeps the original capitalisation, and I don't think you can create pages with the same name but different capitalisation with retain page name case (and if someone did for some reason, I think it's an acceptable sacrifice to not be able to customise them separately).

Let me know what you think! I would propose adding docs for this to the customisation page which I can do if you're happy.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -49,8 +49,8 @@ def test_slugigy():
         slugify("Multiple    Spaces", keep_slashes=True) == "multiple-spaces"
     )
     assert (
-        slugify("With Trailing Slash", keep_slashes=True)
-        == "with-trailing-slash"
+        slugify("With Trailing Slash/", keep_slashes=True)
+        == "with-trailing-slash/"
     )
 
 


deseven:
Nice, I was thinking about something similar.

However, one concern: what about special chars that we allow in page names? Specifically `'"&<>`.

freddyheppell:
Turns out that's all handled fine. Jinja automatically encodes them e.g. the page title `'lots' & "lots" of <special chars>` is output as `data-page-path="&#39;lots&#39; &amp; &#34;lots&#34; of &lt;special chars&gt;"`.

Targeting that with CSS is just:

```css
[data-page-path="'lots' & \"lots\" of <special chars>"] {}
```

deseven:
Yeah, you're right, I thought there will be problems with CSS selectors in that case, but it indeed works as intended in all major browsers:
```html
<html>
<head>
  <style>
    [data-page-path="'lots' & \"lots\" of <special chars>"] {
      background-color: red;
    }
  </style>
</head>
<body>
  <div data-page-path="&#39;lots&#39; &amp; &#34;lots&#34; of &lt;special chars&gt;">test</div>
</body>
</html>
```

redimp:
Hey @freddyheppell, I like this idea. This is great for customizing.

TIL that css selectors can deal with all these characters but it does not feel right, maybe because `data-page-path="&#39;lots&#39; &amp; &#34;lots&#34; of &lt;special chars&gt;"` triggers an OCD ;) 

I would add `slugify()` as jinja filter and use `{{ pagepath | slugify }}` instead of `{{ pagepath.lower()}}`

https://github.com/redimp/otterwiki/blob/bed5cc551920cf8eef14dc82c10704eb5cc80863/otterwiki/util.py#L48-L52

freddyheppell:
Yeah it's not pretty, although it already appears like this a few times such as:

```html
<meta property="og:title" content="&#39;lots&#39; &amp; &#34;lots&#34; of &lt;special chars&gt;" />
<link rel="canonical" href="[http://localhost:8080/&#39;lots&#39;%20&amp;%20%22lots%22%20of%20%3Cspecial%20chars%3E](view-source:http://localhost:8080/'lots'%20&%20%22lots%22%20of%20%3Cspecial%20chars%3E)" />
<title>&#39;lots&#39; &amp; &#34;lots&#34; of &lt;special chars&gt; &ndash; An Otter Wiki</title>
```

And of course that's just how it's escaped in the raw HTML, anywhere you actually see or use it (e.g. the dev tools inspector) it gets decoded back to normal.

Although I agree in some ways slugify would be better, it will remove the slashes. It feels more intuitive to do `[data-page-path^="foo/"]` than `[data-page-path^="foo"]`, which would also match a page named `foobar`. I suppose I could add an option into slugify to support leaving slashes? But I don't want to complicate things too much for a relatively simple feature!

redimp:
> It's not the most elegant solution, but it already exists in places like [..].

Yes, but not in a custom CSS setting/file.

> While slugifying would be cleaner (removing slashes), I think the current approach is more intuitive for cases like `[data-page-path^="foo/"]` compared to `[data-page-path^="foo"]`, which could incorrectly match pages like "foobar".  I could potentially add an option to `slugify` to preserve slashes, but I'm hesitant to overcomplicate a relatively simple feature.

I agree that the slashes should be preserved. Could you please review my implementation to ensure it still meets the requirements?

otterwiki/server.py
```diff
@@ -267,6 +267,17 @@ def format_datetime(value: datetime.datetime, format="medium") -> str:
...
+def slugify(s, keep_slashes=True):
+    """
+    Example usage:
+
+    {{ pagepath | slugify(keep_slashes=True) }}.
+    """
+
+    return otterwiki.util.slugify(s, keep_slashes=keep_slashes)
+
+
```
freddyheppell: By the way if you weren't aware you can just assign the slugify function into the dict of filters, instead of having to write the wrapper like this
https://flask.palletsprojects.com/en/stable/templating/#registering-filters

freddyheppell (commented):
Yeah looks great! Just a couple of small queries

tests/test_util.py
```diff
@@ -39,6 +39,19 @@ def test_slugigy():
...
+        slugify("Some Random Sub/with/A page", keep_slashes=True)
+        == "some-random-sub/with/a-page"
+    )
+    assert (
+        slugify("Multiple    Spaces", keep_slashes=True) == "multiple-spaces"
+    )
+    assert (
+        slugify("With Trailing Slash", keep_slashes=True)
+        == "with-trailing-slash"
+    )
```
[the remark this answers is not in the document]
redimp: Obviously, thanks for checking!

otterwiki/server.py
redimp: Yes, I thought since decorators are used for the other filters, I stick to this method.

redimp:
Released in [v2.16.0](https://github.com/redimp/otterwiki/releases/tag/v2.16.0).

PR #360, opened 2026-01-23
Fix lowercase GitHub-style alerts

emielwiedijk:
GitHub-style alerts in Markdown are case-insensitive. For example, an inline warning in the style of

```markdown
> [!warning]
> Body of the warning
```

GitHub rendering:
> [!warning]
> Body of the warning

is supported on GitHub. While Otterwiki detects this as a warning, the `[!warning]` command is not removed from the rendering of the box, because one regex was not case-insensitive.

Current Otterwiki rendering:

Incorrect rendering of [!warning]

This PR fixes this bug and adds a test that checks whether the command was removed from the rendering.

### DIFF
--- a/otterwiki/renderer_plugins.py
+++ b/otterwiki/renderer_plugins.py
@@ -524,7 +524,8 @@ class mistunePluginAlerts:
     }
     TYPES_WITH_PIPES = "|".join(TYPE_ICONS.keys())
     ALERT_LEADING = re.compile(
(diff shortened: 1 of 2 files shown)


redimp:
Looks good to me, thank you @aimileus!

redimp:
Released in [v2.17.0](https://github.com/redimp/otterwiki/releases/tag/v2.16.0).

Issue #354, opened 2026-01-18
Is it possible to make the "Page Index" the default page on the wiki?


crash5:
I would like to replace the empty "Home" page with the index page. (the "A - Z" page)

redimp:
Hey @crash5, what shows up on the index is currently not configurable.

I see the use-case where this can be useful.

deseven:
Hm, I like the idea. Maybe a setting that would allow you to choose between any root page (with `Home.md` being default) and root page index? This would give enough flexibility for a lot of use-cases.

If this approach looks good I'm willing to do a PR for that later.

redimp:
Thank you for picking this up, @deseven.

The use case I though about was having e.g. `README.md` as `/` so that it fits what your github repository shows first.

redimp:
Released in [v2.17.0](https://github.com/redimp/otterwiki/releases/tag/v2.17.0).

PR #365, opened 2026-01-27
Custom home page

deseven:
Adds a new preference "Home Page" (`HOME_PAGE` for env)  to the "Application Preferences" page and allows you to pick from 3 options:
- Default (Home.md like it is now)
- Root index
- Custom page (by defining any path)

Should cover #354

### DIFF
--- a/otterwiki/preferences.py
+++ b/otterwiki/preferences.py
@@ -177,26 +177,15 @@ def handle_app_preferences(form):
     ]:
         _update_preference(checkbox.upper(), form.get(checkbox, "False"))
 
-    home_page_mode = form.get("home_page_mode", "default")
-    if home_page_mode == "custom":
-        custom_page = form.get("home_page_custom", "").strip()
-        if not custom_page:
-            toast(
-                "Custom home page path is required when 'Custom page' is selected.",
-                "error",
-            )
-            return redirect(url_for("admin"))
-        if custom_page.endswith(".md"):
-            toast(
-                "Custom page path should not include the .md extension.",
-                "error",
-            )
-            return redirect(url_for("admin"))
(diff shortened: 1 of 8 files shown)


deseven:
Added a correct `HOME_PAGE` initialization for cases when we start a new wiki with `HOME_PAGE` being set.

Also now I noticed one thing I overlooked - this approach would prevent user from picking `default.md` or `page_index.md` as home pages. Not sure how big of a problem this might be, maybe I should just make these values more unique, like `_default` and `_page_index`? Or with some prefix, like `wiki:default` and `wiki:page_index`?

redimp:
Hey @deseven, thanks for your work on this feature. I would like to suggest a slightly different approach: if `HOME_PAGE` is empty handle it like it would be "Home". This matches the current behavior.

If it's not empty use `HOME_PAGE` as default for `Page.view()`

If `HOME_PAGE` starts with `/-/` just forward to this address, works for `/-/index` and e.g. `/-/changelog`.

What do you think?

deseven:
So no magic values whatsoever, basically just default (no value) or any path (either to a page in the repo or to a special page), do I get the idea right?

redimp:
Yes, a simple stupid approach.

deseven:
Yeeeeah, I guess I'm a bit rusty after my vacation, not even sure why I overcomplicated it that much :)

This should work better. Also added a simple QoL feature to handle the slashes so the user could write any page name (`some/page` vs `/some/page/`) without thinking about the exact format.

redimp:
Awesome. Will review in detail and test later this week.

redimp:
Released in [v2.17.0](https://github.com/redimp/otterwiki/releases/tag/v2.17.0).

PR #369, opened 2026-01-29
Plugin hook for WikiLinks rendering and example plugin

deseven:
Adds a new hook for WikiLinks and a plugin that marks non-existing pages with red (MediaWiki style).


redimp:
Released in [v2.17.0](https://github.com/redimp/otterwiki/releases/tag/v2.17.0).

PR #373, opened 2026-02-02
Redirect not logged in users to the login page where applicable

### LINKED ISSUE #372: Custom homepage breaks the login page
To reproduce

1. Enable /-/index as default home page
2. Log out
3. Go to http://mypage.com
4. It redirects to http://mypage.com/-/index and throws a 403 error

Looks like the change in #354 introduced this.

deseven:
Fixes #372

### DIFF
--- a/otterwiki/pageindex.py
+++ b/otterwiki/pageindex.py
@@ -7,6 +7,7 @@ from urllib.parse import unquote
 
 from flask import (
     abort,
+    redirect,
     render_template,
     url_for,
 )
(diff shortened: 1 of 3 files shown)


redimp (approved):
Looks good to me. Having proper messages might be a good idea, e.g. in Page.view()

I will rethink how this is handled, but for now lets fix the issue.

redimp:
Fix released in [v2.17.1](https://github.com/redimp/otterwiki/releases/tag/v2.17.1).

deseven:
I'd probably wrap in into a single helper function that accepts redirect url + optional reason for redirect, but passing a message could be problematic. It should either be a part of a URL we're redirecting to, or we need to do internal redirects (display login form on the page that requires the user to be logged in). Both solutions could work.

PR #374, opened 2026-02-02
Fixed broken syntax page

deseven:
The `/-/syntax` page cannot be loaded after changes introduced in f4af9a8.

The `pagepath` was undefined even before in that case, but using `lower()` on it was returning an empty string, so this wasn't a problem. Now it is :)


redimp:
Fix released in [v2.17.1](https://github.com/redimp/otterwiki/releases/tag/v2.17.1).

PR #376, opened 2026-02-03
Add page path lowercasing logic to redlink plugin

### LINKED ISSUE #375: Red links plugin marks all red in some environments
I've broken this out into a new issue because it's a bit more complex than it just not working.

When I run v2.17.1 with the redlinks plugin installed in my local dev env (macOS 26.2), the plugin functions as expected and only marks nonexistent links as red. However, when I run it on my server (dockerised, using the slim image variant), all links are marked as red.

If I have a page called "My Page", it's stored on disk as "my page.md". If I modify the plugin code to log some stuff and load a page containing the markdown `[[My Page]]`:

```py
page_file = (
    page_path if page_path.endswith('.md') else f"{page_path}.md"
)

print(page_file)
print(self.storage.exists(page_file.lower()))
```

I get the output:
```
My Page.md
True
```

So the issue is that it's checking for the existence of the file "My Page.md", which doesn't exist, instead of "my page.md" which would. In both envs, the "Retain page case" setting is set to its default of off.

I don't entirely understand the different behaviour of capitalisation cross-platform and how that interacts with otter wiki and the page case setting, but I'm assuming it's a bit more complicated than just always lowercasing the page check.

freddyheppell:
Fixes #375 by lowercasing the page path unless case retention is enabled.

I did try just importing `otterwiki.helper.get_filename()` which contains this logic, the `.md` extension handling and slash cleaning, but it causes a circular import error as the helpers try and import `otterwiki.app`.

### DIFF
--- a/docs/plugin_examples/plugin_redlinks/otterwiki_redlinks.py
+++ b/docs/plugin_examples/plugin_redlinks/otterwiki_redlinks.py
@@ -19,6 +19,7 @@ class RedLinks:
     @hookimpl
     def setup(self, app, db, storage) -> None:
         """Initialize the plugin with access to core objects."""
+        self.app = app
         self.storage = storage
 
     @hookimpl
@@ -43,6 +44,13 @@ class RedLinks:
         page_path = page_path.strip('/')
         if not page_path:
             return wikilink_html
+
+        # Unless case is retained, convert to lowercase for storage check
+        page_path = (
+            page_path
+            if self.app.config["RETAIN_PAGE_NAME_CASE"]
+            else page_path.lower()
+        )
 
         # Check both with and without .md extension for directories
         page_file = (


deseven:
LGTM, I think it's okay for a plugin to handle such things internally. Thanks for the fix!

GoldenFlame0:
Trying to run the code locally after putting in the patch by hand, fails on startup with a complaint about the conditional. I'm adding screenshots just in case I'm an idiot and made a typo.

The conditional is also a little funky, in my mind I'd go "if self.app.config(...)" and then if that's true, set the path to lower case. That way you skip an assignment if you don't need it, and you don't need the else in there. Although I'm not much of a python guy, your way might be a more idiomatic way to do it. I know re-assigning strings is weird because they're immutable. EDIT: you could possibly move this check up to the initial assignment.

image

image

redimp:
@GoldenFlame0 you have a typo in your code:

```
   if self.app.config=["RETAIN_PAGE_NAME_CASE"]
                     ^
                     | remove this
```

Netx time pleeeeaaaase post the code and not a screenshot of the code. Thank you!

GoldenFlame0:
Haha, _knew_ it was a bad idea to try and test this at 2am. I'll make sure to post the text directly next time, I was being a bit lazy by using a screenshot.

deseven:
Maybe you just got confused by this patch a bit :) For people who are not that familiar with Python this could look weird.

This block reads like that: leave `page_path` as is if `RETAIN_PAGE_NAME_CASE` is true, otherwise lowercase the `page_path`.

It's a pretty normal construction for Python, although I agree that
```python
if not self.app.config["RETAIN_PAGE_NAME_CASE"]:
    page_path = page_path.lower()
```
would probably be better here. But it's more a stylistic choice than anything else.

GoldenFlame0:
Oh yeah, it was a nitpick, but I thought I'd point it out.

Can confirm the changes work correctly on my server as well, although we're way past that point now lol.

PR #381, opened 2026-02-09
Fix confusing editor behaviour on URL pasting

### LINKED ISSUE #380: URL pasting large blocks (testing.otterwiki.com)
This is in regards to a feature on the testing page: https://testing.otterwiki.com/Testing/URL%20Pasting

When I have a URL in my clipboard and I select a word on the edit page, then hit paste, the editor automatically creates a Markdown link: the selected word becomes the link text, and the clipboard URL becomes the target.

While this is sometimes useful, it often leads to behavior I don’t expect.

My usual workflow is: I copy something to the clipboard, select some text I want to replace, and press paste. Depending on what exactly I select, this sometimes works, but sometimes the “smart” Markdown link behavior kicks in instead.

A few observations:
- If I select a plain URL and paste, the special behavior understandably should trigger.
- However, if I select, for example, an entire list and then paste, I really don’t expect the whole list to be wrapped in `[ ]` instead of being removed and replaced.
- This paste behavior seems to be triggered as soon as the clipboard content starts with `http(s)`. So if I copy a large block of text that just happens to begin with a URL, the result becomes very confusing and messy when I paste.

In the image example i took the whole three point list and copied it, starting from `https://oterwiki.com` and pasted it onto the whole list again.

Image

This might be considered an edge case, but it makes paste operations sometimes do something different from what I expect.

Possible improvement: support something like Ctrl+Shift+V as “paste without special features” (similar to “paste without formatting” in other tools), so I can reliably replace selected text with the raw clipboard content.

deseven:
Fixes #380 as discussed.

I also added pasted URL trimming just in case. We can introduce even more rules, but I think this is a reasonable compromise between flexibility and consistency.

### DIFF
--- a/otterwiki/static/js/otterwiki.js
+++ b/otterwiki/static/js/otterwiki.js
@@ -1182,6 +1182,9 @@ var otterwiki_editor = {
         // if an url is pasted on a selected text, check if it should be
         // replaced with a markdown link []()
         if (cm_editor.getSelection().length > 0) {
+            // don't activate link behavior if selected text contains line breaks
+            if (selected_text.includes('\n')) { return false; }
+
             state = otterwiki_editor._getState();
             // we don't mess with existing tokens of these kinds
             if (state.image || state.link || state.code || state.url) { return false; }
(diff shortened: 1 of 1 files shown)


redimp:
Fix released in [v2.17.2](https://github.com/redimp/otterwiki/releases/tag/v2.17.2).

Issue #221, opened 2025-03-19
Links to Pages that Reference Page


kaleb112494:
Thought of this while adding pages. Our wiki references pages from others quite frequently. Is there a possibility of adding a side section that lists pages where the current page is referenced? 

e.g. you're on /software/rufus, On the right side below "On This Page", there's a section that lists these pages that reference /software/rufus

/hardware/reloads/manual
/software/automation/sop for automation fails
/sops

This would help when creating new pages that link elsewhere. Not a huge deal, but would be one less thing to remember when creating documentation.

redimp:
Hey @kaleb112494, thanks for opening up this request. I have to think about the use cases. 

My first thought us that this will mostly pollute the sidebar(s) and will not add much value.

To clarify: Do you want this displayed while editing or while viewing the page?

Or should this be a "tool" like e.g. `Blame` that could be a drop-down from the "Three dots menu"?

kaleb112494:
@redimp Sorry for the delayed response. In the scenario that I have in my head, this menu would be presented while viewing, not while editing. It would be more of a tool to find out more information about the subject that's on this page. Here's an image of what I was conceptualizing. My apologies for the terrible photoshop job.

https://imgur.com/a/jB3qkdU

I suppose I may be using OtterWiki unlike most people, but the documentation I'm creating is separating what the subject is from how it is used, and more so in some cases.

FreeSoftwareServers:
I love this idea basically like a sources/reference section on a wiki page. You know when you snowball looking into things and you end up clicking on links to read more about related things lol That's the use case. Sort of like how I ended up on this issue page...

Almeus:
I think this is typically called "backlinks" or "bidirectional links" by eg Mediawiki and Obsidian.  It's the "what links here" on the Wikipedia sidebar.

PR #386, opened 2026-02-12
Plugin system improvements

deseven:
I was originally trying to just write a plugin to cover #221, but ran into some issues with the way pluggy works and ended up fixing them as well, so this PR is a bit too broad, sorry. Hopefully it won't cause any issues with your ongoing work on the embeddings, but please do tell me if it will and I can fix them later.

Changes:
1. Added 3 new plugin hooks - to inject html into left sidebar, to inject html into right sidebar and to react on repository changes.
2. Fixed the usage of default argument values in both hookspecs and hookimpls since they are [not supported by pluggy](https://github.com/pytest-dev/pluggy/issues/442).
3. Added basic descriptions for all arguments in hookspecs.
4. Added an example plugin that should solve #221.

### DIFF
--- a/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
+++ b/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python3
+
+"""
+Referencing Pages Plugin for An Otter Wiki
+
+This plugin demonstrates the template_html_sidebar_right_inject and repository_changed hooks.
+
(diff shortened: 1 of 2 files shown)


otterwiki/templates/layout.html
```diff
@@ -213,8 +213,10 @@
         <div class="content-wrapper" id="content-wrapper">
 
             <div class="container-fluid">
+{# Capture extra_nav content once to avoid re-rendering the block multiple times #}
```
redimp: Nice.

PR #387, opened 2026-02-12
feat(docker): add label requried for renovate to detect the changelog

m-wrzesien:
At the moment when [Renovate](https://docs.renovatebot.com/) (tool for updating dependencies) doesn't know where to look for otterwiki changelog.

With this change, it will be able to fetch them from the release page, as described [here](https://docs.renovatebot.com/modules/datasource/docker/#description).

Example of the PR before this change:
image

Example of the PR with image that has this label:
image

Also thanks for creating such nice software :D


redimp:
Simple enough. Thanks for this @m-wrzesien!

PR #388, opened 2026-02-14
Broken WikiLinks checker for Housekeeping

### LINKED ISSUE #385: [Feature Request] Housekeeping - Find broken links
A plugin was recently added that shows broken links in red which works great!

It would be very useful to me to add a way to find broken links in all pages, similar to the current feature in the Housekeeping pages to find empty pages. When we rename or move a page, all links to it become broken and it is not easy to find them.

I realise that the red links feature was implemented as a plugin, but it seems like such a basic feature of a wiki app that I think it might a good idea to add this as a feature in the base app and not force everyone to install a plugin for it.

Thanks for an Otterwiki, it's awesome.

deseven:
Fixes #385, also added housekeeping tests.

### DIFF
--- a/otterwiki/templates/tools/housekeeping_brokenwikilinks.html
+++ b/otterwiki/templates/tools/housekeeping_brokenwikilinks.html
@@ -28,7 +28,7 @@
             </td>
             <td style="white-space: normal; word-wrap: break-word;">
                 {% for link in broken_links %}
-                    <span>[[{{link}}]]</span>{% if not loop.last %}, {% endif %}
+                    <a href="{{ url_for('view', path=link) }}" target="_blank">{{link}}</a>{% if not loop.last %}, {% endif %}
                 {% endfor %}
             </td>
             <td style="width:1px;text-align:right;padding-right:.5rem;">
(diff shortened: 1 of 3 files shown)


redimp:
Looks god so far and an extra hooray for adding tests. 

I would like to recommend two improvements: 

(1) In the list of pages with broken links, add a column with the first four(?) pages linked.

(2) The inverted result would also be useful: "Most Wanted Pages". A list of pages that exist as broken links with a list of the first four(?) pages referring to each of the pages.

deseven:
Done and done. Also added basic sorting.

I didn't limit the number of links/pages, instead I allowed word-wrapping.

redimp:
Nice! some minor things to be improve:

### Pages without broken links

- List of links in "Broken links" should be unique
- Relative links are not detected correct, e.g. `[[../Home]]`
- Displaying the correct link would enable the user to just create the page, maybe display this instead of the syntax snippet.

image

### Most wanted pages

- List of pages in "Found in" should be unique
- `#fragments` should not be displayed, just the pagename

image

deseven:
Done. I'm not 100% sure of the relative links logic, though...

Well, this got far beyond the original issue and to be honest I personally don't see much use in it (a simple list of pages with broken links is all I need), hopefully someone would :) Still, as something to consider in the future, it might make sense to do these checks automatically (weekly?) and send reports to the admin users.

PR #393, opened 2026-02-20
Current page highlight on sidebar

ryanwhowe:
This is a feature to highlight the current page on the sidebarmenu.  This will change the render of the current page from an `<a>` tag to a `<span>` tag.

When there are many pages on the same level having the highlight on the sidebarmenu is helpful. I could not think of any reason to keep the link on the sidebarmenu clickable and navigable when on the page itself.

I reused the theme color from the top breadcrumb render.  The full path is used to determine what to highlight so there should be no issues with overlaping names of pages.

This is the render of a top level page with no children
image

This is the render of a parent page
image

This is the render of a child page
image

Open to any feedback.

### DIFF
--- a/otterwiki/static/css/partials/sidebar.css
+++ b/otterwiki/static/css/partials/sidebar.css
@@ -210,6 +210,10 @@ a.sidebar-toc-6 { padding-left: calc(7 * var(--spacing) - var(--radius) - 13px);
     color: rgba(255, 255, 255, 1);
 }
 
+.dark-mode .sidebarmenu span {
+    color: var(--dm-link-text-color);
+}
+
 .sidebar-link {
     padding-left: 1rem;
 }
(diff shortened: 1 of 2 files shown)


redimp:
A great idea @ryanwhowe, thank you for this PR!

crash5:
Is it only for the dark mode? I don't see any change on the light theme.

ryanwhowe:
I'll be completely honest, I did not even realize that there was a light mode.  I can recreate this in demo, it was a missed css rule.  I will create a new PR to address this.

ryanwhowe:
Thanks for the catch @crash5, I have a PR in to add the highlight to light mode in #396 .

crash5:
thank you! :)

redimp:
Thanks for reporting @crash5 and thanks for immediately implementing @ryanwhowe!

PR #394, opened 2026-02-21
feature: fancy statistics (without extra dependencies)

Jocho-Smith:
The Idea is to have a dashboard in the settings to review the number of commits made by users.

Here is how it looks like now:

image

On my machine, it passes all the tests.

Note:
- I'm unhappy of the location of the js in dashboard.html. Like I say in the commit message, such a script is likely better located in wiki.html, however I can't estimate the implications this would have on the entire otterwiki. Do I oversee an obvious way to improve on this?
- Since otterwiki handels Mermaid in Markdown differently then what I do here, the dark mode gets ignored and the colors are always optimized for light mode.

Happy to hear any feedback :)

### DIFF
--- a/otterwiki/statistics.py
+++ b/otterwiki/statistics.py
@@ -5,8 +5,9 @@ from otterwiki.gitstorage import GitStorage
 
 
 class StatisticsService:
-    def __init__(self, repository_path: str):
+    def __init__(self, repository_path: str, top_limit: int = 10):
         self.storage = GitStorage(repository_path)
+        self.top_limit = top_limit
 
     def _aggregate_commits(
         self,
@@ -25,7 +26,20 @@ class StatisticsService:
             author = entry["author_name"]
             stats[author] = stats.get(author, 0) + 1
 
-        return sorted(stats.items(), key=lambda x: x[1], reverse=True)
+        sorted_stats = sorted(stats.items(), key=lambda x: x[1], reverse=True)
+
+        # If within limit, return directly
+        if len(sorted_stats) <= self.top_limit:
+            return sorted_stats
+
+        top_entries = sorted_stats[: self.top_limit]
+        remainder = sorted_stats[self.top_limit :]
+
+        others_count = sum(count for _, count in remainder)
+
+        top_entries.append(("Others", others_count))
+
+        return top_entries
 
     def commit_statistics(
         self,
@@ -33,6 +47,7 @@ class StatisticsService:
     ) -> List[Tuple[str, int]]:
         """
         Returns aggregated commit counts per author.
+        Only top contributors are shown; remainder grouped as 'Others'.
         """
         log_entries = self.storage.log()
         return self._aggregate_commits(log_entries, since_days)
--- a/otterwiki/templates/admin/statistics.html
+++ b/otterwiki/templates/admin/statistics.html
@@ -16,7 +16,7 @@
 
 <div class="mermaid">
 pie showData
-    title All Commits test
+    title All Commits
 {% for author, count in all_commits %}
     "{{ author }}" : {{ count }}
 {% endfor %}


Jocho-Smith:
An ideas I just had: I noticed that after ~12 users/entries the pie chart starts to look a bit messy. I could modify the function to only show the top 10 users and concatenate the remainder users into on entry.

otterwiki/views.py
```diff
@@ -248,6 +249,57 @@ def admin_mail_preferences():
         return otterwiki.preferences.handle_mail_preferences(request.form)
 
 
+@app.route("/-/admin/dashboard")
```
redimp: We want to prevent to have any complex logic in the `otterwiki.views` Module. Please move the logic into e.g. `otterwiki.statistics`.

otterwiki/views.py
```diff
@@ -248,6 +249,57 @@ def admin_mail_preferences():
...
+            "Last Month", ["--format=%an", "--since=1.months"]
+        )
+    except RuntimeError as e:
+        return str(e), 500
+
+    # Merge both blocks
+    markdown_content = "\n\n".join([all_commits_block, last_3_months_block])
+
+    # Render Markdown
+    htmlcontent, _, library_requirements = render.markdown(markdown_content)
```
redimp: I'm torn here ... It is a nice idea to use the renderer for displaying internal statistics module. On the other hand, just adding the mermaid plain to the template skips the need for using the renderer here.

Jocho-Smith:
Thank you for this extensive feedback! I'll put in some time to implement the required changes.

> I'm thinking about what else might be nice to see when looking at statistics/graphs of a wiki:
>  -  a code frequency graph with additions and deletions

I don't understand what exactly you mean by that. Can you share a picture of it?


> - an activity heat map, showing when changes happened

Actually I implemented this some time ago, however I don't see how to do that with Mermaid. What package would you be fine to distribute it within otterwiki for this Statistics feature? I've seen Flask+Plotly some times.

Here is how I did it with seaborn:
image

Jocho-Smith:
I implemented all suggested changes now. Thanks again for exactly pinpointing what needed to be done.

I also implemented the above mentioned cap at 10 users (which can be changed later, potentially with the settings.cfg)

Also the 30 days is just a value I picked now it can also be changed to other values, like I did here:

image

Jocho-Smith:
So what do you think how should we continue from here?

Do you want to take this simple mermaid statistics feature as it is - maybe as an "experimental feature" (as in 'Repository Management') - or should we dive into more sophisticated packages (like Plotly)?

redimp:
This looks much better thanks for the effort.

## Are Pie Charts the best option?

I checked how these charts look like for different projects and came to the conclusion that Pie Charts are not the best way for displaying this. In most projects the number of commits is very unbalanced a lot of people only have a small number of commits. I think a bar chart would be better suited, like github does it, see <https://github.com/redimp/otterwiki/pulse>:

image

I gave it a try and this looks readable to me:

image

Code:

````markdown
```mermaid
---
config:
  xyChart:
    showDataLabel: true
    width: 500
    height: 400
  themeVariables:
    plotColorPalette: '#1890ff'
---
xychart
title "February 21, 2026 – February 28, 2026"
y-axis "Commits"
x-axis ["redimp", "deseven", "ryanwhowe", "Prodeguerriero","gokomer"]
bar [9,3,2,2,1]
```
````

Note: I [updated mermaid to 11.12.3](https://github.com/redimp/otterwiki/commit/fbd44f91aea3b1d51005a8b3f6fb2d1594a5c3f7) to make this work.

What do you think?

## Other options for diagrams

For now I don't want to add more dependencies if not really necessary. Adding e.g. `plotly` opens a special can of worms. But I'm open for suggestions/PRs with working examples (layout, theme, light/dark mode).

## More diagrams to display:

Heapmap: Yes what you created with seaborn is exactly what I had in mind. This needs no library, you can build this with a html table ;)

## Layout

Another thing to optimize is the layout of the diagrams, the two diagrams shouuld be displayed side by side if the screen size allows it (so everything wider than mobile screens).

Jocho-Smith:
Hey @redimp 

> Note: I [updated mermaid to 11.12.3](https://github.com/redimp/otterwiki/commit/fbd44f91aea3b1d51005a8b3f6fb2d1594a5c3f7) to make this work.
> 
> What do you think?

I love it! This was actually my initial plan, I just accepted the current mermaid version and went with what is available. Obviously one can just update mermaid xD

> This needs no library, you can build this with a html table ;)

Cool idea, I'll work on that.

> Another thing to optimize is the layout of the diagrams, the two diagrams shouuld be displayed side by side if the screen size allows it (so everything wider than mobile screens).

This could probably also be done with just html/css - I think flexbox does the job. I'll try that out.


### plan for this PR
So what would you suggest is the objective of this PR? xychart + html heatmap + flexbox = merge? Or should we discus this more? 

Sorry btw. for crashing in with a PR, without opening an Issue for discussion first.

redimp:
> So what would you suggest is the objective of this PR? xychart + html heatmap + flexbox = merge? Or should we discus this more?

In general I merge PRs when I have nothing to ask for anymore ;)

I'm way more picky for completely new features: In this case, when this statistics page looks nice and I think users of An Otter Wiki can get something useful from it, I will merge it.

Jocho-Smith:
Ok. Then I guess we will see where this goes. 

For now I'll work on the things you suggested ✌️

PR #396, opened 2026-02-23
Added a light mode sidebar css rule.

ryanwhowe:
Reported in #393 there was a missing css rule for the `span` render in light mode to highlight the current page highlight.
Added a missed css rule for highlighting the new `span` render for the current page on the sidebar

Render in light mode
image

This addition did not effect the dark mode render
image

### DIFF
--- a/otterwiki/static/css/partials/sidebar.css
+++ b/otterwiki/static/css/partials/sidebar.css
@@ -203,6 +203,10 @@ a.sidebar-toc-6 { padding-left: calc(7 * var(--spacing) - var(--radius) - 13px);
     border-color: transparent;
 }
 
+.sidebarmenu span {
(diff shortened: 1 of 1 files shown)


ryanwhowe:
There is still a minor top padding difference between the links and the span that was not as apparent when it was a small list as it is to me in a large list.   When there is a long list under the selected page the "jump" is very apparent.  I'll try and get another PR in some time in the near future ... css is one of my less favorite things 😆 
image
image
It is hard to see, and it really takes watching the jump when there are a dozen or two pages on the same level.  But after about 6-7 items the jump is VERY visible.

crash5:
What do you think about leaving it as a link and just changing the style? (if I get it right, the jump is because the span tag only)

otter-span-jump

redimp:

PR #397, opened 2026-02-24
Clean html improvements

deseven:
This changes the logic of `clean_html()` from being a blacklist to being a whitelist instead, since I found multiple ways to do harmful stuff otherwise.

Also, the user is now allowed to explicitly allow anything using the new config parameter `RENDERER_HTML_WHITELIST`, which follows a (hopefully) simple format of `tag1[attr1 attr2],tag2,tag3[attr3]`, for example: 
```bash
RENDERER_HTML_WHITELIST=iframe[src width height],svg,button
```

Please check out the list of default allowed tags and attributes, see if something needs to be added or removed.

### DIFF
--- a/tests/test_renderer.py
+++ b/tests/test_renderer.py
@@ -261,6 +261,184 @@ def test_clean_html_script():
     )
 
 
+def test_clean_html_xss_vectors():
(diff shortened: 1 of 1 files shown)


redimp:

PR #398, opened 2026-02-26
Adjust font size for form control in CSS

gokomer:
On ios clicking form fields zooms in. This pr override halfmoon's form control font size to prevent mobile zoom.


redimp:

PR #408, opened 2026-03-10
UI: fix vertical jumps in sidebar when changing pages

tvanas:
This small PR fixes vertical jumps in the sidebar when changing pages.

Before:
https://github.com/user-attachments/assets/75ff46f6-eb3c-47b1-bd61-11fe51a04650

After:
https://github.com/user-attachments/assets/1a9a3b43-66f7-4b1e-afd1-742ead81d93d

PR #410, opened 2026-03-11
Bug/Sidebar span line height fix

ryanwhowe:
Added the missing CSS rule to make the line-height the same for `<a>` and `<span>` entries in the `sidebar-menu`.

Without this rule, the line height defaulted to the `.sidebarmenu li` rule, which sets the line-height to 1, which would resolve to the standard font size for the line-height.

I know it is hard to see in the images, but this was the render before the update
original-span-spacing

This is the render after the update
updated-span-spacing

It is only a 3-pixel difference on my screen, but the menu moving during navigation was not an ideal experience.  Sorry, it took 3 PRs to get this feature in correctly. 

related to #393 and #396

### DIFF
--- a/otterwiki/static/css/partials/sidebar.css
+++ b/otterwiki/static/css/partials/sidebar.css
@@ -186,7 +186,7 @@ a.sidebar-toc-6 { padding-left: calc(7 * var(--spacing) - var(--radius) - 13px);
     margin-bottom: .5rem;
 }
 
-.sidebar-menu a {
+.sidebar-menu a, .sidebar-menu span {
     line-height: 2rem;
 }
 


redimp:
Hey @ryanwhowe, thank you for this PR. I'm currently on the road will check on this on the weekend. Can you check https://github.com/redimp/otterwiki/pull/408 I think you are solving the same issue.

ryanwhowe:
Thank you @redimp, you are correct, we are solving the same issue both here and in #408, we are, however, solving it in completely different ways.

@tvanas is converting the `<span>` back to a `<a>` tag.  My PR just adds the missing CSS rule for the `line-height` for the `<span>` entries.  I think that either method will work.  I originally had changed the `<a>` tags to `<span>` tags explicitly to make them not navigable.  I could not see a use case where you would want to click a link to go to a page that you are currently on.  I was even considering adding something similar to the breadcrumbs on the top of the page in the future.

If you go with the #408 solution, I think that the template can be reduced.  The `if` `else` logic was added in #393 due to having to render different elements.  We could reduce it back down to a single line with a conditional attribute

i.e.
```html 
<a href="/{{value["path"]}}"{% if pagepath and pagepath.lower() == value.path.lower() %} aria-current="page"{% endif %}>{{value["header"]}}</a>                    
```

I prefer my solution, but I'm not married to it; it was a preference thing, saw it as 2 birds with 1 stone when I initially implemented it.
1. Highlight the current page
2. Make the current page not navigable

tvanas:
What a coincidence :)

Thanks for the comparison, @ryanwhowe. I took the `<a aria-current="page">` approach because it felt like a clean way to indicate the current page while removing the jump. I kept the template close to the existing style, though I agree it could be simplified inline if that approach is preferred.

That said, keeping the current page as a `<span>` is a reasonable choice too. I don't have a strong preference as long as the jump is fixed cleanly.

ryanwhowe:
Hope your travels are going well @redimp , I think @tvanas and I are in accord.  Both approaches are absolutely valid and it is just a judgement call (sorry we couldn't just dismiss one out of hand for ya 😝 ).  Don't want to speak for @tvanas but I think we both are fine with whichever direction you'd like to go in ... we just want that sidebar jump to 🛑 🤣

redimp:
Hey @ryanwhowe and @tvanas, thank you both for the additional thoughts and comments.

> I could not see a use case where you would want to click a link to go to a page that you are currently on. I was even considering adding something similar to the breadcrumbs on the top of the page in the future.

A good example for why user experience design is hard, at least for me. I always try to give the user as much options as possible with the limitation that it keeps the visuals clean. In this example a `<a>` does not hurt someone who never clicks on it, but does someone who wants to make use of this. 

For example: I use the link in the breadcrumbs to copy the link when I want to share the link to the page. I could add a "share" button that copies the link to the clipboard or opens up a submenu with different options but this would clutter the page head.

I go with @tvanas solution and follow your advice to reduce this back to a single line. Will tag the both of you in the Release notes and changelog.

I appreciate the effort to make An Otter Wiki a better tool for everyone a lot.

PR #426, opened 2026-03-24
Migrate to Mistune 3.2.0

onovy:
Quickly tested manually, unit tests passed


redimp:
Hey @onovy, thanks for this PR. From a first glance this looks pretty good.

I've found only one issue for now, the FancyBlocks are broken, e.g.

```markdown

:::info
# An info block

with some content.
:::

and some text.

```

renders as

image

and should render as

image

onovy:
should be fixed now, thanks

PR #431, opened 2026-03-27
fix: replace eval() in editor preview with direct DataTables reinit

onovy:
Remove the eval(data.preview_js) call in the editor live-preview fetch handler and replace it with a targeted simpleDatatables reinit loop over tables matching the known id prefix. Also remove preview_js from the server-side preview JSON response, eliminating the eval attack surface at the API level. DataTables is already unconditionally loaded in the editor page via force_load_libraries, so the reinit works without any additional script loading.

### DIFF
--- a/otterwiki/templates/editor.html
+++ b/otterwiki/templates/editor.html
@@ -670,16 +670,11 @@ thx to https://github.com/sparksuite/simplemde-markdown-editor/issues/328#issuec
                 preview_block.innerHTML = data.preview_content;
                 sidebar_toc.innerHTML = data.preview_toc;
                 extranav_toc.innerHTML = data.preview_toc;
-                if (data.preview_js)
-                {
-                    try {
-                        eval(data.preview_js);
-                    }
-                    catch(e) {
-                        console.log("Error", e.stack);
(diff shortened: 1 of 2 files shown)


redimp:
Hey @onovy, thank for this PR. I think it's a good idea to re-initialise DataTables in the preview pane without eval().

Two issues here:

(1) When you change the options of the `{{DataTable}}` in the editor and hit `Preview` the change will not be reflected by just re-initializing the datatable.

(2) We can not remove calling the `renderer_javascript` hook entirely. This is a plugin hook which is supposed to be not just used by the `DatatableEmbedding`.

onovy:
ad 2) it's not used elsewhere in Otterwiki code so it could be required by third-party/user's plugins, right?

redimp:
(2) This plugin hook is pretty new, I expect more plugins to come that might use this hook.

onovy:
ok, eval is not best approach but i understand why you using it here. Attack surface is minimal because eval code is from plugin. I don't think it's worth to do special fix just for data tables and keep eval for other plugins, thus closing.

PR #432, opened 2026-03-27
fix: use secrets.choice instead of random.choice for password generation

onovy:
But TBH: This generator is never called. So this is probably for some future feature?

### DIFF
--- a/otterwiki/util.py
+++ b/otterwiki/util.py
@@ -5,7 +5,7 @@ import datetime
 import mimetypes
 import os.path
 import pathlib
-import random
+import secrets
 import re
 import regex
 import string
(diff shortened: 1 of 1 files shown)


redimp:
Hey @onovy, I think this is a solid improvement.

`random_password()` is called in `create_user()` when no password is given:

https://github.com/redimp/otterwiki/blob/89f5de3d9a64abadc94ccbde178e5ae3aabf2de3/otterwiki/auth.py#L232-L237

onovy:
and ```create_user``` is never called with ```password=None``` :)

redimp:
True.

redimp:
Will merge anyway, better to have a more secure password generation and not use it, than have a weak one and not use it.

redimp:

PR #433, opened 2026-03-27
fix: correct always-false guard and empty log message in check_passwo…

onovy:
…rd_hash_backport


redimp:

PR #434, opened 2026-03-27
fix(search): prevent ReDoS via user-supplied regex with timeout

onovy:
Replace stdlib re with the regex library which supports a timeout parameter. Pass timeout=5 to all match calls that use user-supplied patterns (Search class). Handle regex.TimeoutError gracefully so a catastrophic regex returns no results instead of hanging the process. Bump dependency regex 2024.11.6 → 2026.2.28.

### DIFF
--- a/otterwiki/wiki.py
+++ b/otterwiki/wiki.py
@@ -2,7 +2,7 @@
 # vim: set et ts=8 sts=4 sw=4 ai:
 
 import os
-import re
+import regex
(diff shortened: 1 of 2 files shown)


redimp:
Hey @onovy, looks good to me. Will add some toasts() to inform the user if the search times out.

redimp (approved):

redimp:

Issue #146, opened 2024-10-09
Documentation on using gunicorn and uv run


simonw:
The https://otterwiki.com/Installation instructions currently lean towards "production" installations - Docker, Kubernetes, uwsgi.

I wanted to try it out on my laptop with the least possible steps - it turns out this worked for me:

```bash
git clone https://github.com/redimp/otterwiki.git
cd otterwiki

mkdir -p app-data/repository
git init app-data/repository

echo "REPOSITORY='${PWD}/app-data/repository'" >> settings.cfg
echo "SQLALCHEMY_DATABASE_URI='sqlite:///${PWD}/app-data/db.sqlite'" >> settings.cfg
echo "SECRET_KEY='$(echo $RANDOM | md5sum | head -c 16)'" >> settings.cfg

uv run --with gunicorn gunicorn --bind 127.0.0.1:8080 otterwiki.server:app
```
This `uv run` command automatically creates a virtual environment, installs the `pyproject.toml` dependencies, installs `gunicorn` as well (`--with gunicorn`) and starts that running the server.

simonw:
It would be cool if the [otterwiki PyPI package](https://pypi.org/project/otterwiki) could be run like this:

```bash
pip install otterwiki
otterwiki \
  --repository app-data/repository \
  --sqlite app-data/db.sqlite \
  --secret-key secret1 \
  --port 8080
```
That would make it even easier for people to just try it out.

e12e:
Nice. I assume your first snippet assumes `pip install uv` or equivalent?

redimp:
Hey @simonw, I like to idea of running An Otter Wiki via a shell command. At least for testing in a non-production environment this would be very useful! With using not just flask but running a wsgi server (for example uv as you suggested) even in production environments.

I'll take a look at how this can be implemented.

sfinnie:
small suggestion: gunicorn doesn't run on windows.  [waitress](https://pypi.org/project/waitress/) is pure python, supported by Flask, and runs across all platforms. Using instead would prevent issues for windows users & keep the process simple per @simonw 's suggestion (`pip install` & run).

PR #438, opened 2026-03-29
Enable running locally without Docker

sfinnie:
A partial solution to [Issue 146](https://github.com/redimp/otterwiki/issues/146).  Enables running locally, without Docker, using [waitress](https://pypi.org/project/waitress/) as the web server.  Waitress used instead of [gunicorn](https://gunicorn.org/) as it supports windows too.

Includes convenience scripts (bash & powershell) for running that take care of config and initialising the content git repo. To satisfy [#146] fully, these could be ported into the otterwiki `server` instead.

**NOTE**: `run_local.sh` uses bash.  I believe it's still installed on MacOS (as of v26 Tahoe) but unable to verify as I don't have access to a recent Mac environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ On Linux/MacOS/BSD and other UNIX-like systems:
 On Windows:
 
 ```powershell
-run_local.ps1
+.\run_local.ps1
 ```
 
 4. Open your browser at [http://localhost:8080](http://localhost:8080)


onovy:
maybe i'm missing something here, but running `make run` just runs locally installation without docker. It's using Flash internal web server (not for production use).

sfinnie:
> maybe i'm missing something here, but running `make run` just runs locally installation without docker. It's using Flash internal web server (not for production use).

Thanks for the feedback.  There are 3 small advantages this PR brings:
1. `make` isn't installed on windows.  
2. It automates configuration, including initialising the content git repo, simplifying use.
3. It supports "local production" use.

I use An Otter Wiki for personal notes at home and work.  It's not accessible to anyone else, just the local desktop.  I share content via the git repo.  I can't use the Docker-based config at work since Docker Desktop requires a paid license for commercial use on windows.

I realise that's perhaps niche and the 3 enhancements combined may not be sufficiently valuable to merge; no problem if so.  I submitted the PR because #146 got a positive response.

Thanks.

redimp:
Hi @sfinnie, thanks for submitting this PR! I'm currently evaluating the best approach to address this issue.

I appreciate you bringing up `waitress`. It's interesting that it's so lightweight.

@onovy, I've noticed that using `flask` can sometimes have issues even with a single user, particularly when dealing with pages containing many attachments.

Ideally, I'd like to explore solutions like these:

```bash
$ pip install 'otterwiki[uwsgi]'
$ python3 -m otterwiki --repository ./repository --sqlite ./db.sqlite
```

and / or

```bash
$ pip install 'otterwiki[waitress]'
$ otterwiki --repository ./repository --sqlite ./db.sqlite
```

Potentially all variants: Depending on which WSGI server is installed and both as module to call `python3 -m otterwiki` and as script `otterwiki`.

sfinnie:
Thanks for the response @redimp , makes sense.

Running the module directly would be better than shell scripts.  A small suggestion: perhaps provide defaults for the content repo and DB that can be overridden (on the command line like your examples / via environment).  That would make initial use even easier since some users might not know good values for the repo dir and db name.

sfinnie:
Closing given discussion above - thanks for comments.

PR #440, opened 2026-03-30
docs: add AGENTS.md knowledge base for AI coding agents

onovy:


redimp:
Hey @onovy, for now I don't want an `AGENTS.md` in this repository.

The LLM/Coding Assistant/Agents World is moving to fast to keep up to date for which agent handles which prompt how. I barely keep up with the PRs, bugs and feature requests currently happening and I don't want to invite low effort "I just added the feature X with N lines of code over M files." PRs.

Don't get me wrong, I'm not opposed to people using agents. These are tools as any other. And they help a lot. For example with reviewing PRs ;) But I don't want to deal with PRs that took a tiny fraction to write in comparison with the time needed to be reviewed.

onovy:
I understand. TBH all my PR was written with assistance by agents :) I don't have problem with keeping this AGENTS.md locally, but it could be usefull for others contributors.

redimp:
Happy to come back to this as soon as there is some convergence in the agentic world.

deseven:
A couple of points on the proposed `AGENTS.md` in general:

1. It's just too broad as it tries to cover a lot of ground, but at the same time too specific since it's mentioning even the specific dependency versions. This will require constant maintenance and updates with little benefit.

2. I saw no trouble with any modern models understanding what it is and how it works, since the project is not that complicated anyway. Usually some (minor) troubles come from things that are not that common or depend on the environment, specifically:
   - the usage of make
   - venv
   - tests and coverage
   - settings.cfg and repo initialization

   In my opinion these are the only four things that should be explained to the model in advance.

PR #445, opened 2026-04-01
fix(auth): prevent account enumeration via generic recovery messages

onovy:


onovy:
> Sometimes I feel like I don't have enough imagination to figure out how to cause harm.

Don’t be in a bad mood. These are very low-severity issues, and overall, this product is in good shape! Nice work and thanks!

redimp:

Issue #462, opened 2026-04-19
Outdated frontend dependencies, considering contributing


fidergo-stephane-gourichon:
Hi,

Testing otterwiki, I and a few friends are considering adopting it. We noticed the frontend dependencies are fairly old and that upgrading them means manually replacing files. Here's what we found:

- CodeMirror 5.65.15 (latest: 6.0.2)
- Font Awesome Free 5.15.4 (latest: 7.2.0)
- MathJax 3.2.0 (latest: 4.1.1)
- Mermaid 11.6.0 (latest: 11.14.0)
- htmx 2.0.6 (latest: 2.0.8)
- Sortable.js 1.15.3 (latest: 1.15.7)

We're considering helping, e.g. by upgrading the packages and/or making future upgrades easier, preferably in a pythonic way (although unsure how applicable this is to non-Python packages).

Anyway, before proposing anything specific I'd rather ask first: you know best what fits how you maintain the project. What would be your preferences?

redimp:
Hey @fidergo-stephane-gourichon, thank you for reaching out and your interest in An Otter Wiki.

About dependencies, I have an (until now) unwritten strategy: If it isn't broken and doesn't bring features I need, I keep the versions stable. Also, some upgrades come with a lot of work but not much gain. Let me go into the details:

- `CodeMirror`: I worked on moving to CodeMirror 6 for some time and gave up eventually. The code (not rebased in a long time) lives in [feature/cm6-proof-of-concept](https://github.com/redimp/otterwiki/tree/feature/cm6-proof-of-concept). Also check the discussion in #105. Please go ahead and take a look—you will find mobile users very happy with it. Checklist:
    -   Theme working decent, including light and dark mode.
    -   Syntax Highlighting supported, including code blocks (check [otterwiki.com/Examples/Syntax%20Highlighting](https://otterwiki.com/Examples/Syntax%20Highlighting)).
    -   All toolbar functionality supported (e.g., modifying markdown tables).
    -   Right Sidebar tooling supported (Attachments and Wikilinks).

- `Font Awesome`: I didn't miss anything from Version 5, and I wasn't keen on some of the alternatives in Version 6. Please go ahead and try a mockup for Version 7; if it looks good, I'm easily convinced.

- `MathJax & Mermaid`: Go ahead. Please test and verify manually that everything works.

- `htmx & sortable`: Go ahead. These should be straightforward and easier to test.

[...]

PR #465, opened 2026-04-19
Upgrade/htmx sortablejs

fidergo-stephane-gourichon:
This follows #462 

> htmx & sortable: Go ahead. These should be straightforward and easier to test.

htmx: I guess the thing would not work at all if broken.
sortable: given it is use in sidebar_preferences.html and I have the same result as on https://demo.otterwiki.com/-/admin/sidebar_preferences I guess it's okay?

I followed the existing pattern of copying the minified version available from public source. 
Perhaps another pattern like in 3869aa63 would make easier to review.


redimp (approved):
Looks good to me. SHA-256 sums match unpkg CDN.

redimp:
Released in [v2.20.3](https://github.com/redimp/otterwiki/releases/tag/v2.20.3).

PR #476, opened 2026-05-03
Security check report for housekeeping

deseven:
Added a security check section for Housekeeping.

Implemented checks:
- Anonymous users can edit the wiki
- Anonymous users can upload files
- Server name not set
- CSP header missing
- HTTPS not enabled
- Fully open registrations
- Plugins are being used
- Custom HTML/CSS/JS is being used
- Custom HTML tags/attrs whitelist
- Missing of misconfigured reverse proxy

Most of the checks are performed on the backend side with a couple running on the frontend instead since we need to see the situation from the outside. Backend checks are solidly organized and easily extensible I think, while the frontend ones are a bit hacky, so if someone has any better ideas please share (not a blocker, though - could be improved later).

### DIFF
--- a/otterwiki/security_check.py
+++ b/otterwiki/security_check.py
@@ -1,0 +1,428 @@
+#!/usr/bin/env python
+# vim: set et ts=8 sts=4 sw=4 ai:
+
+"""
+otterwiki.security_check
+
+Extensible security check framework for Otterwiki.
+Each check is a function registered via the @register_backend_check decorator
+that returns a SecurityCheckResult (or a list of) if an issue is found, None otherwise.
+"""
+
+import os
+import re
+import ipaddress
+
+from flask import request
+from otterwiki.server import app
+from otterwiki.plugins import plugin_manager
+from otterwiki.util import empty
+
+
+class SecurityCheckResult:
+    """Represents a single security check finding."""
+
+    def __init__(self, issue, description, severity):
+        self.issue = issue
+        self.description = description
+        self.severity = severity
+
+    def to_dict(self):
+        return {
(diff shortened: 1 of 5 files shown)


redimp:
Thanks for the PR, this is a really nice addition.

I went ahead and pushed a few follow-up commits with the changes I would otherwise have asked for in review:

- Gate the endpoint and the panel on ADMIN, so the config and
  active plugin list aren't exposed to read-only accounts.
- Reuse halfmoon's badge classes (with two new badge-warning /
  badge-notice variants in otterwiki.css) instead of inline
  styling, so dark mode and theming work out of the box.
- Switch _has_html_content to BeautifulSoup since bs4 is already
  a hard dependency.
- Drop the REAL_IP_FROM branches from the reverse proxy check;
  that variable only applies to the bundled docker/nginx setup
  and was misleading on a generic install.
- Extend SecurityCheckResult with a passed flag and render a
  collapsible "Passed checks (N)" details section underneath the
  issues table, so admins also see what the wiki got right.
- Downgrade the anonymous write/upload and HTTPS checks from
  CRITICAL to MEDIUM when the request comes in over a loopback
  host, with a note in the description explaining the downgrade.

Happy to iterate on any of these if you'd prefer a different
shape.

deseven:
Makes sense to me and sorry about not checking for admin, I keep forgetting that this page is available for regular users too because of the drafts functionality. And no idea why I haven't added the admin check for the endpoint.

Security check with fatal security flaws 👌

redimp:
Released in [v2.20.5](https://github.com/redimp/otterwiki/releases/tag/v2.20.5).

PR #477, opened 2026-05-03
Make 'APPROVED' new default for write and attachment access

deseven:
Addressing previous security concerns, this makes `APPROVED` new default for `WRITE_ACCESS` and `ATTACHMENT_ACCESS`.

This change broke a lot of tests, so I made some adjustments, adding a distinction between logged in and anonymous users where needed, so we test against our new default configuration instead of something that is more convenient. Any feedback on that is appreciated!


redimp:
Thanks for the PR. As discussed, I think this is a useful change to the defaults: it prevents accidents while still allowing users to run wikis that can be edited anonymously.

Will review in detail and merge in the next days.

deseven:
Fixed conflicts + a couple of extra tests.

PR #483, opened 2026-05-10
Forward onto wiki page after authenticating

Harriotskelli:
If you go to a wiki page and aren't authenticated you get redirected to login.
But Login doesn't forward back to that wiki page from history like it does for settings and other static pages. 

This adds that functionality to these by putting in the Next parameter.
If the URL in Next isn't valid, then we're put back to the home page anyway.

### DIFF
--- a/otterwiki/pageindex.py
+++ b/otterwiki/pageindex.py
@@ -9,6 +9,7 @@ from flask import (
     abort,
     redirect,
     render_template,
+    request,
     url_for,
 )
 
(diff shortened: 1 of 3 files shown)


redimp:
Hey @Harriotskelli, thank you for bringing this up and adding a PR to fix it. I went a bit further and fixed it also for all other permission gates.

redimp:
Released in [v2.20.5](https://github.com/redimp/otterwiki/releases/tag/v2.20.5).

PR #498, opened 2026-05-29
fix referecing pages plugin

ribbal:
fixes one issue:

- WIKILINK_STYLE setting is not taken account of
- ~the referencing links do not display the full path of the linking page~

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,30 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Flask: Run OtterWiki",
+            "type": "debugpy",
+            "request": "launch",
+            "module": "flask",
+            "args": [
+                "run",
+                "--port",
+                "8080",
+                "--no-reload"
+            ],
+            "env": {
+                "FLASK_ENV": "development",
+                "FLASK_DEBUG": "True",
+                "FLASK_APP": "otterwiki.server",
+                "OTTERWIKI_SETTINGS": "../settings.cfg"
+            },
+            "console": "integratedTerminal",
+            "justMyCode": false,
+            "serverReadyAction": {
+                "pattern": "Running on http://127.0.0.1:8080",
+                "uriFormat": "http://localhost:8080",
+                "action": "openExternally"
+            }
+        }
+    ]
+}
--- a/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
+++ b/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
@@ -39,6 +39,8 @@ class ReferencingPages:
         self.app = app
         self.storage = storage
 
+        self.WIKILINK_STYLE = app.config.get("WIKILINK_STYLE", "")
+
         self._build_reference_index()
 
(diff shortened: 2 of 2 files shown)


ribbal:
fixes #499

deseven:
Hi, thank you for this PR!

Not sure if the second change is needed, `{pagename_full}` is already used in the title that could be seen if you hover over the referencing page link. If we're going to print full page paths this will look quite ugly as each link could span multiple lines. I'd say in general it's not a common practice to have multiple identical page names that are located in different subdirectories while referencing a single page.

If you really think this could be useful, maybe add an ENV parameter (like `PLUGIN_REFERENCINGPAGES_FULL_PATH`) with false as a default?

ribbal:
@deseven, thanks for the feedback.  I'm new to Otter Wiki, so I'll take on board your comments.  I've reverted the changes for the second issue.

If it proves to be a common problem (for other users) it can be revisited - or better still, convert the referencing pages to a collapsible structure (like the Page Index in the left hand menu, perhaps).

redimp:
Hey @ribbal, thanks for your PR! Thanks @deseven for reviewing.

Question unrelated to the issue but the PR. I would prefer not to add `.vscode/launch.json`  or any other devenvironment config to the repo, but I could be conviced (since I put vim modelines into the files ;) ) to do so. Here the question:

Is the `.vscode/launch.json` valid for any vscode or would this work on your machine?

ribbal:
Hi @redimp, I was expecting this pushback regarding the `.vscode/launch.json`.  It was a conscious decision to add it to the source code repository.

I am new to python development, and spent a lot of time trying to get the development environment to work in VS Code.  The `launch.json` was the end result that allowed me to run VS Code so that it would start the server and automatically launch a browser with the correct url.

I would like to keep this file, for the following reasons:

- it may help other (VS Code) developers
- it will safeguard the contents of the file (that I spent a lot of time authoring)

Of course, if a developer doesn't use VS Code, there will be no impact by the inclusion of `.vscode/launch.json`.

To address your specific question above (_if I understand it correctly_), the contents of `launch.json` are valid for anyone's VS Code.  The settings are **not** specific to my machine.

Please let me know if you would still like me to remove the file?  In this case, I will need to safeguard the contents locally.

deseven:
I've tested it. Apart from requiring the `debugpy` extension (which I guess you should have if you want to debug Python apps via Code) there's nothing that is environment-specific, it pretty much mimics `make debug`.

However, some recommendations for @ribbal in general:
- it's a good practice to clearly state in the commit message what these changes are about, e.g. `added vscode launch configuration`
- the PRs should only have the commits relevant to the described issue/enhancement, adding some extra stuff without explicitly stating what is it and what it is for is considered to be a bad move
- your git user is configured incorrectly, so the commits are not tied to your GH profile, set `git config --global user.name` and `git config --global user.email`, you currently have it set to `user <user@vm-debian13-tech.local>`

None of this is too important for this particular case, since the aren't many changes and your commits can be squashed, but if you're willing to make contributions in the future this might be something to consider.

ribbal:
Thanks for the feedback @deseven.  I'll take your comments on board for the future.

Issue #215, opened 2025-03-10
Support for Opening Links in a New Tab


kyle-mckay:
Part of what I’m using Otterwiki for, is to create wiki pages with direct links to service providers and other external resources. However, I’d like the ability to open these links in a new tab instead of the current window.  

According to resources like [GeekforGeeks](https://www.geeksforgeeks.org/how-to-create-links-with-target_blank-in-markdown/), some Markdown editors and applications support the `target="_blank"` attribute using syntax like this:  

```markdown
Here is an [example](https://example.com){:target="_blank"} link
```

Currently, in Otterwiki, this syntax is not supported. When I try to use it, the `{:target="_blank"}` part is rendered as plain text instead of being applied to the link.  

![Image](https://github.com/user-attachments/assets/02b7f941-dc33-4f58-a414-7a2d98a2acef)

Does Otterwiki support the ability to open links in a new tab, either through the `target="_blank"` attribute or a similar mechanism? If not, this could be a potential enhancement that would greatly improve the usability of external links in wiki pages in my own opinion.

redimp:
Hey @kyle-mckay, thanks for opening up this issue. Currently this is not supported.

I never read anywhere of a markdown extension that is mentioned in the GeekforGeeks article. There is also not mentioned which markdown parsers support this feature, I guess at least their very own parser? Some stackoverflow posting mention that kramdown supports it, but I can not find it in their docs: https://kramdown.gettalong.org/syntax.html

In general I have the impression that `target="_blank"` have gone out of fashion, since most browsers support "Opening in a new tab" via the middle mouse button .. nevertheless I will check how different markdown flavors support this and will think about adding this feature!

FreeSoftwareServers:
I mean I'm a power user so I always just hold CTRL and click which opens a new tab, but I tend to always prefer a new tab most regular users don't know that trick...

ribbal:
Hey @kyle-mckay, would it satisfy your use case if **ALL external links** were **automatically** opened in a new tab?

So, no need to create any additional markdown, and no need to implement any additional markdown parsing.

@redimp, would you support this approach?  If so, I can look at creating a PR - it would of course be controlled by a new ('_Open external links in a new tab_') config setting.

I would definitely like to see this functionality being implemented.

ryanwhowe:
[...]

PR #507, opened 2026-06-15
open external links in new tab

ribbal:
open external links in new tab, controlled by new config option: "Open external links in a new tab."

Addresses: #215 (perhaps)

### DIFF
--- a/otterwiki/renderer.py
+++ b/otterwiki/renderer.py
@@ -3,6 +3,7 @@
 
 import re
 import mistune
+import urllib.parse
 from bs4 import BeautifulSoup
 from markupsafe import Markup, escape
 from mistune.plugins.formatting import strikethrough as plugin_strikethrough
@@ -347,7 +348,7 @@ class OtterwikiMdRenderer(mistune.HTMLRenderer):
         if title:
             attrs.append('title="{}"'.format(mistune.escape(title)))
 
-        if open_in_new_tab and not link.startswith('/'):
+        if open_in_new_tab and self.is_external_link(link):
             attrs.append('class=external')
             attrs.append('target=_blank')
 
@@ -362,6 +363,11 @@ class OtterwikiMdRenderer(mistune.HTMLRenderer):
             self.env.get('page'),
         )
         return processed_html
+
+    def is_external_link(self, link):
+        parsed_link = urllib.parse.urlparse(link)
+
+        return parsed_link.scheme in ('http', 'https')
 
     def codespan(self, text):
         if len(text) >= 3 and text.startswith("$") and text.endswith("$"):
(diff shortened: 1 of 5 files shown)


otterwiki/renderer.py
```diff
@@ -332,15 +332,26 @@ def image(self, text, url="", title=None):
     def link(self, text, url=None, title=None):
         if empty(text):
             text = url
-        # escape url, title
+
+        # escape url
         link = mistune.escape_url(self.safe_url(url))
 
+        open_in_new_tab = self.env.get("config", {}).get(
+            "OPEN_LINKS_IN_NEW_TAB", ""
```
redimp: Defaulting to `False` instead of `""` looks cleaner (to me).

otterwiki/renderer.py
```diff
@@ -332,15 +332,26 @@ def image(self, text, url="", title=None):
...
+
         if title:
-            link_html = '<a href="{}" title="{}">{}</a>'.format(
-                link, mistune.escape(title), text
-            )
-        else:
-            link_html = '<a href="{}">{}</a>'.format(link, text)
+            attrs.append('title="{}"'.format(mistune.escape(title)))
+
+        if open_in_new_tab and not link.startswith('/'):
```
redimp: This check is not reliable enough. For example these all are rendered as external links:

```markdown
- [Some Page](SomePage)
- [Section](#anchor)
- [Mail](mailto:mail@example.com)
- <https://url-to-your-wiki.tld/Example> (see below)
```

My suggestion: Parse `link` with `urllib.parse.urlparse`, check the scheme and only if the `scheme` matches http(s) consider it an external link.

If a user adds the external url to a page of the wiki on purpose `<https://url-to-your-wiki.tld/Example>` I think it's okay to mark it as external link.

otterwiki/static/css/otterwiki.css
```diff
@@ -571,3 +571,16 @@ mark {
     background-color: var(--gray-color);
     border-color: var(--gray-color);
 }
+
+.external[target="_blank"]::after {
```
redimp: No need for `[target="_blank"]`. I would use

`a.external::after {`

redimp (changes requested):
Hey @ribbal, thanks for this PR. Please address the requested changes.

ribbal:
@redimp, thanks for the review.  I've made the requested changes.  Please could you check?

otterwiki/renderer.py
ribbal: Agreed - Fixed.

otterwiki/renderer.py
ribbal: Agreed - Fixed (as per the feedback comments).

otterwiki/static/css/otterwiki.css
ribbal: Agreed - Fixed.

redimp (approved):
Looks good to me.

redimp:

Issue #228, opened 2025-03-27
Improve Page Rename


redimp:
As discussed in #210, renaming a page could make use of some features:

- a) the links to attachments on the page itself (and other pages, using the same attachments)
- b) the links to the page form other pages
- c) corner case: When renaming attachments, the links shoudl be updated, see https://github.com/redimp/otterwiki/issues/65

These should be optional when renaming pages, so that you have to check a box, when you want links to be auto updated. There are cases, where users rename pages on a regular basis and don't want the links to be updated.

skirmess:
I took a stab at this as an **example plugin**: #508.

It updates WikiLinks pointing to a page when that page is renamed, via the `page_renamed` hook. Anchors (`[[Page#section]]`) and explicit `[[title|link]]` titles are preserved, and matching goes through `get_filename()` so it's case-/slash-/`.md`-insensitive.

Scope is intentionally limited to WikiLinks for now (plain Markdown links and attachments/#65 are not covered). Since a per-rename checkbox isn't reachable from the current plugin API, it uses an `UPDATE_LINKS_ON_RENAME` config flag (default on) so it can be disabled without uninstalling — e.g. when baked into a Docker image.

redimp:
Released in [v2.23.0](https://github.com/redimp/otterwiki/releases/tag/v2.23.0)

Issue #65, opened 2023-12-10
Renaming attachments should updated links in pages.


redimp:
When a user renames an attachment, the existing links in the pages should be updated with the same commit.

Reported in #60.

PR #508, opened 2026-06-16
docs: add updatelinksonrename example plugin

skirmess:
This adds an example plugin under `docs/plugin_examples/` that updates WikiLinks when a page is renamed, implementing the request from #210 / #228.

### What it does
On rename, the plugin hooks into `page_renamed` and rewrites every WikiLink (`[[...]]`) across the wiki that points to the old page so it points to the new name. Matching uses `get_filename()` (the same resolver Otter Wiki uses), so it is case-, slash- and `.md`-insensitive, and `#anchor`s and explicit `[[title|link]]` titles are preserved.

### Scope / limitations
- Only WikiLinks are rewritten — matching what the bundled `referencingpages` example can discover. Plain Markdown links (`[text](/Page)`) are intentionally left untouched.
- Exact match of the renamed page only; child pages and attachment links (#65) are not touched (child pages aren't moved by a rename, so links to them stay valid).
- A per-rename checkbox (as floated in #210) isn't exposable via the current plugin API, so an `UPDATE_LINKS_ON_RENAME` config flag (default on) lets you disable it without uninstalling — handy when the plugin is baked into a Docker image.

### Tests
Adds `tests/test_plugin_updatelinksonrename.py` covering the rewrite (anchor + title preserved, unrelated links untouched) and the opt-out flag. I noticed the example plugins currently have no tests, so happy to drop it or move things around if you'd rather keep them test-free.

### DIFF
--- a/docs/plugin_examples/plugin_updatelinksonrename/otterwiki_updatelinksonrename.py
+++ b/docs/plugin_examples/plugin_updatelinksonrename/otterwiki_updatelinksonrename.py
@@ -1,0 +1,163 @@
+#!/usr/bin/env python3
+
+"""
+Update Links on Rename Plugin for An Otter Wiki
+
+When a page is renamed this plugin rewrites all WikiLinks (``[[...]]``)
+in the wiki that point to the old page so that they point to the new
+page name instead. This implements the request from
+https://github.com/redimp/otterwiki/discussions/210 (issue #228).
+
+It demonstrates the ``page_renamed`` hook and shows how a plugin can
+read, modify and commit page content through the storage API.
+
+Scope / limitations:
+- Only WikiLinks (``[[Page]]``, ``[[Title|Page]]``) are updated, matching
+  what the bundled "referencingpages" example can already discover. Plain
+  Markdown links (``[text](/Page)``) are intentionally left untouched.
+- Matching is exact for the renamed page. Child pages and attachment links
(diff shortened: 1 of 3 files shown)


redimp:
Hey @skirmess, thanks for this PR. Will test, review, and come back to you.

ribbal:
My thoughts on this is that it should be part of the Otter Wiki **core** functionality.  What do you think @redimp, @skirmess?

Or, am I fundamentally missing something?

I'm sure @skirmess's good work could still be re-used.

skirmess:
Yes, that would probably make sense to have it in core (it's more a bug fix than a new feature). But I'm not a Python dev, I can't even really read it - and I'm not going to use AI to edit someone else's software if I don't understand what the change does. :-)

That's why I asked Claude to write a plugin. Because I wanted that feature now.

You can implement it in core, with or without using part of this MRs code. Or add the plugin to the repo. Or close the MR. All fine with me. Claude wrote it, I just tested it on my installation.

ribbal:
Implemented this page rename functionality into the core otter wiki code: #530.

redimp:
Since this PR is obsolete due to #530 which added this as a feature to the core, I close this PR. Added credit in the [v2.23.0](https://github.com/redimp/otterwiki/releases/tag/v2.23.0) release notes, though.

PR #509, opened 2026-06-16
fix: match anchored WikiLinks in referencingpages example

skirmess:
The `referencingpages` example plugin indexes WikiLink targets verbatim. A link that carries a `#anchor` (e.g. `[[Page#section]]`) or a leading slash was stored as `page#section`, which never matches the anchor-less page path used when looking up backlinks for the sidebar. The result: such backlinks are silently missing.

I hit this on a real page — a referrer linking via `[[Some/Page#hello]]` did not show up under "Referencing pages", while a plain `[[Some/Page]]` from another page did.

### Fix
Strip the `#anchor` and a leading slash before normalizing the target, and skip pure in-page anchor links like `[[#section]]`.

### Tests
Adds `tests/test_plugin_referencingpages.py`, which asserts that both an anchored and a plain referrer show up as backlinks (the anchored one fails without this fix), and that a pure `[[#section]]` link creates no backlink. (Heads-up: the example plugins didn't have tests yet, so this introduces a small importlib-based loader for the docs/ plugin — easy to drop if you'd rather keep them test-free.)

### DIFF
--- a/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
+++ b/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
@@ -99,8 +99,14 @@ class ReferencingPages:
 
             # Process each WikiLink
             for match in matches:
-                # Clean up the page name
-                target_page = match.strip()
+                # Clean up the page name: drop any #anchor and a leading
+                # slash so the target matches the page path used for lookup
+                # (e.g. [[Page#section]] and [[/Page]] both point to "Page").
+                target_page = match.split('#', 1)[0].strip().lstrip('/')
+
+                # Skip pure anchor links like [[#section]]
+                if not target_page:
+                    continue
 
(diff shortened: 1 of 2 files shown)


ribbal:
@skirmess, I believe that this issue has been resolved by: #527.  Therefore, can this PR be closed?

redimp:
Hey @skirmess, thank you for this PR. I'm sorry, this somehow was lost on my TODO list.

> I believe that this issue has been resolved by: #527.

Mostly, yes. #527 does now strip `#anchor` from WikiLink targets and skip pure `[[#section]]` links, so the anchored-link case this PR was about is resolved. I ran this PR's two regression tests against current `main` and both pass.

One gap remains, though: the **leading-slash** case. #527 didn't add the `.lstrip('/')` part, so a referrer using `[[/Page]]` still produces no backlink, even though the renderer treats `[[/Page]]` and `[[Page]]` as the same page.

I will cherry pick the missing parts.

PR #525, opened 2026-07-08
Add possibility to override the roles for the read/write/upload/admin permission

Getslow6:
Hi!

 > Note: this PR is a work in progress and not tested yet, but I'm submitting it already as I would like to check if this functionality is something that could be added. If this functionality would be accepted, I could help update documentation and testing the functionality etc as well.

When setting up Otterwiki, I found that I needed to add the READ/WRITE/UPLOAD/ADMIN permissions to my authentication layer in order to provide the correct permissions. However, in my case I'm using roles instead of direct permissions, such as 'member, moderator, admin'.

In this PR I've added a few new environment variables:

    AUTH_ROLES_READ="READ"
    AUTH_ROLES_WRITE="WRITE"
    AUTH_ROLES_UPLOAD="UPLOAD"
    AUTH_ROLES_ADMIN="ADMIN"

The default values for the roles are the same as the values for the different permissions now, so it is backward compatible. 
With this PR, it would be possible to override the roles as such:

    AUTH_ROLES_READ="member, moderator, admin"
    AUTH_ROLES_WRITE="moderator, admin"
    AUTH_ROLES_UPLOAD="moderator, admin"
    AUTH_ROLES_ADMIN="admin"

This way the roles in any authentication middleware can be mapped to the permissions inside Otterwiki. 
I'm curious to hear your feedback!


redimp:
Hey @Getslow6, thank you for this PR.  I get the idea of what you want to achieve and think this can be useful. 

Will test and review later this week (hopefully, my weekend is packed).

otterwiki/auth.py
```diff
@@ -662,13 +683,13 @@ def request_loader(self, req):
         if self._email_header not in req.headers:
             app.logger.error(f"Missing header: {self._email_header}")
             return None
-        if self._permissions_header in req.headers:
-            permissions = (
-                req.headers[self._permissions_header].upper().split(',')
+        if self._roles_header in req.headers:
+            roles = (
+                req.headers[self._roles_header].upper().split(',')
             )
```
redimp: There is a whitespace mismatch between the configured roles and the header roles: `_parse_roles` strips each token (`r.strip().upper()`), but the header value is split with a plain `.upper().split(',')`.

With the example config from the PR description (`AUTH_ROLES_ADMIN="admin"`) and a proxy sending `x-otterwiki-permissions: moderator, admin` (space after the comma, which is common), the header tokenizes to `['MODERATOR', ' ADMIN']`, so `' ADMIN'` never matches and the user silently ends up with `{READ, WRITE}` but no `ADMIN`.

Reusing the helper here fixes the mismatch and removes the duplicated parsing:

```suggestion
            roles = self._parse_roles(req.headers[self._roles_header])
```

otterwiki/auth.py
```diff
@@ -662,13 +683,13 @@ def request_loader(self, req):
         if self._email_header not in req.headers:
             app.logger.error(f"Missing header: {self._email_header}")
             return None
-        if self._permissions_header in req.headers:
-            permissions = (
-                req.headers[self._permissions_header].upper().split(',')
+        if self._roles_header in req.headers:
+            roles = (
+                req.headers[self._roles_header].upper().split(',')
```
redimp: Better parse the `re.header` with `_parse_roles()` 

Header roles are not whitespace-stripped, while configured roles are, so any role after a comma+space never matches here.

otterwiki/auth.py
```diff
@@ -677,11 +698,15 @@ def request_loader(self, req):
             return None
         elif empty(name):
             app.logger.warning(
-                f"name header '{self._username_header}' is emptyt"
+                f"name header '{self._username_header}' is empty"
```
redimp, line 697: \*cough\*

otterwiki/auth.py
```diff
@@ -624,18 +624,39 @@ def supported_features(self):
...
+        write_roles: str,
+        upload_roles: str,
+        admin_roles: str,
     ):
         self._username_header = username_header
         self._email_header = email_header
-        self._permissions_header = permissions_header
-        self.logout_link = logout_link
+        self._roles_header = roles_header
+        self._read_roles = self._parse_roles(read_roles)
```
redimp: I think it would be better to turn this into a single mapping 
```python
self._role_map = {
'READ': ..., 'WRITE': ..., 
'UPLOAD': ..., 'ADMIN': ...
}
```
This reduced `_map_roles_to_permissions` to a single loop.

otterwiki/server.py
```diff
@@ -37,6 +37,10 @@
     AUTH_HEADERS_USERNAME="x-otterwiki-name",
     AUTH_HEADERS_EMAIL="x-otterwiki-email",
     AUTH_HEADERS_PERMISSIONS="x-otterwiki-permissions",
+    AUTH_ROLES_READ="READ",
```
redimp: Please remove the trailing whitespaces. I recommend running `black` e.g. via `make black` before committing / pushing.

redimp (commented):
Hey @Getslow6, thanks for submitting this PR. I did a quick read-through and made some comments with my findings.

Will add some tests for `AUTH_METHOD="PROXY_HEADER"` to make sure nothing breaks with this PR.

redimp:
Tests added in https://github.com/redimp/otterwiki/commit/7af1672084af659c6c5b8a564f4f2a29a862d60b .

otterwiki/auth.py
Getslow6: I've applied your suggestion.

otterwiki/server.py
Getslow6: Whitespace removed

otterwiki/auth.py
```diff
@@ -624,18 +624,37 @@ def supported_features(self):
 
 
 class ProxyHeaderAuth:
-    # if logout_link is not provided, hide the logout button
     def __init__(
         self,
-        logout_link=None,
-        username_header: str = 'x-otterwiki-name',
```
Getslow6, line 631: I've removed these defaults, as they are already defined in server.py.

Getslow6:
Hi @redimp, I've implemented your suggested changes. Thanks for the review.
I'm not familiar with testing this, but I can try to do so in the coming weeks. 

For transparency; I've used quite some python when I studied so I can find my way around the code. The PR and way to implement it were mine, but Claude helped me a bit with the changes (which I ofcourse reviewed). I hope that is okay!

redimp:
Thanks for the transparency. I appreciate that. I'm fine with using LLMs/Coding assistance for developing. I use it for reviewing my code.

I draw a line at slop, which I define as low-effort, low-quality digital content generated in high volumes. PR must take more time to write than to read.

Will test and review your changes later this week.

redimp:
Looks all good to me. Merged.

redimp:

PR #527, opened 2026-07-11
enhance referencing pages plugin

ribbal:
This change enhances the referencing pages plugin, in three ways:

1. fix alignment of the page link (see the BEFORE screenshot below)
2. when displaying on a narrow screen, the 'referencing pages' disappears - the fix is to snap it to the left sidebar in the same way as 'on this page'
3. works for _anchor links_ to the current page

image

### DIFF
--- a/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
+++ b/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
@@ -104,7 +104,8 @@ class ReferencingPages:
             # Process each WikiLink
             for match in matches:
                 # Clean up the page name
-                target_page = match.strip()
+                target_page = urllib.parse.unquote(match.strip())
+                target_page = target_page.split("#", 1)[0].strip()
 
                 # Normalize the page path (handle case sensitivity)
                 if not self.app.config.get("RETAIN_PAGE_NAME_CASE", False):
--- a/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
+++ b/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
@@ -90,9 +90,9 @@ class ReferencingPages:
             # WikiLink patterns: [[Page Name]] or [[Page Name|Title]]
             # or:                [[Page Name]] or [[Title|Page Name]]
             wikilink_pattern = (
-                r'\[\[([^\]|]+)(?:\|[^\]]+)?\]\]'
+                r'\[\[(?!#)([^\]|]+)(?:\|[^\]]+)?\]\]'
                 if self.WIKILINK_STYLE.upper() == "LINKTITLE"
-                else r'\[\[(?:[^|\]]+\|)?([^\]]+)\]\]'
+                else r'\[\[(?![^|\]]*\|#)(?!#)(?:[^|\]]+\|)?([^\]]+)\]\]'
             )
             matches = re.findall(wikilink_pattern, content)
 
@@ -106,10 +106,6 @@ class ReferencingPages:
                 target_page = urllib.parse.unquote(match.strip())
                 target_page = target_page.split("#", 1)[0].strip()
                 target_page = self.normalize_page_name(target_page)
-
-                # Handle pure anchor links (e.g. #sub-section)
-                if not target_page:
-                    continue
 
                 # Skip self-references (a page cannot reference itself)
                 if target_page == source_page:
--- a/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
+++ b/docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
@@ -3,17 +3,21 @@
 """
 Referencing Pages Plugin for An Otter Wiki
 
-This plugin demonstrates the template_html_sidebar_right_inject and repository_changed hooks.
+This plugin demonstrates:
+- template_html_sidebar_left_inject
+- template_html_sidebar_right_inject
+- repository_changed hooks
 
 Features:
 - Builds an in-memory index of page references (which pages link to which)
 - Updates the index when repository changes are detected
-- Displays a "This page is referenced by" block in the right sidebar
+- Displays a "Referencing pages" block in the sidebar
   showing pages that have WikiLinks pointing to the current page
 
 This plugin demonstrates:
 1. setup() - Initialize the plugin and build the initial reference index
 2. repository_changed() - Update the index when files change
+3. template_html_sidebar_left_inject() - Add content to the left sidebar
 3. template_html_sidebar_right_inject() - Add content to the right sidebar
 """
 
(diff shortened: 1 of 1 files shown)


ribbal:
This is the AFTER screenshot of the page link alignment:

image

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
```diff
@@ -211,11 +229,12 @@ def template_html_sidebar_right_inject(self, page):
             )
             # Build the HTML for the sidebar block
             # Using the same structure as the "On this page" block
-            html = '''
-    <details class="collapse-panel" open>
-        <summary class="collapse-header">Referencing pages</summary>
+            html = f'''
+    <div id="extranav-toc" {' class="sidebar-toc d-xl-none"' if is_left_sidebar else ''}>
```
redimp: Please pick a different id. `#extranav-toc` is already being used. This might break the editors preview.

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
```diff
@@ -228,10 +247,11 @@ def template_html_sidebar_right_inject(self, page):
                 # URL encode the page path
                 url_path = urllib.parse.quote(pagename_full)
 
-                html += f'            <a href="/{url_path}" class="sidebar-link" title="{pagename_full}">{pagename}</a>\n'
+                html += f'                <a href="/{url_path}" class="sidebar-link sidebar-toc-1 sidebar-active" title="{pagename_full}">{pagename}</a>\n'
```
redimp: I think not every link should have class `sidebar-active`.

redimp (changes requested):
Thanks for improving this plugin! Please address my findigs.

Two other things I stumbled upon:
 
1. Pure anchor links e.g. `[[#section]]` will pollute the index, they should be filtered out.
2. The self-reference check compares the lowercased target against the non-lowercased source. Before this PR that mismatch was latent, but now `[[Home#section]]` inside `Home.md` normalizes to target `home` vs source `Home`, so with the default `RETAIN_PAGE_NAME_CASE=False` the page will list itself under "Referencing pages". Normalize `source_page` the same way before comparing.

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
ribbal: updated to: `backlinks-toc`

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
ribbal: removed `sidebar-active`

ribbal:
Thanks for the review @redimp.

I've resolved the two inline comments above.

Regarding the other points:

1. Pure anchor links have been filtered out
2. `source_page` has been normalised to match `target_path` (_plus a little code tidy up_)

Please could you review the changes?

ribbal:
Just one more thing - Whilst I'm at it, I'd like to rename this demonstration plugin from '**_referencing pages_**' to '**_backlinks_**', which I believe is the more common name for this concept amongst other wikis.

Thoughts?

redimp:
> Just one more thing - Whilst I'm at it, I'd like to rename this demonstration plugin [..]

Naming is hard (at least for me). I would prefer not to change the name to not break existing deployments.

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
```diff
@@ -3,17 +3,21 @@
...
 - Updates the index when repository changes are detected
-- Displays a "This page is referenced by" block in the right sidebar
+- Displays a "Referencing pages" block in the sidebar
   showing pages that have WikiLinks pointing to the current page
 
 This plugin demonstrates:
 1. setup() - Initialize the plugin and build the initial reference index
 2. repository_changed() - Update the index when files change
+3. template_html_sidebar_left_inject() - Add content to the left sidebar
 3. template_html_sidebar_right_inject() - Add content to the right sidebar
```
redimp, line 20: This should be `4.` ;)

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
```diff
@@ -192,30 +207,27 @@ def template_html_sidebar_right_inject(self, page):
...
                 (k, [get_pagename(k), get_pagename(k, full=True)])
                 for k in referencing_pages
             )
             # Build the HTML for the sidebar block
             # Using the same structure as the "On this page" block
-            html = '''
-    <details class="collapse-panel" open>
-        <summary class="collapse-header">Referencing pages</summary>
+            html = f'''
+    <div id="backlinks-toc" {' class="sidebar-toc d-xl-none"' if is_left_sidebar else ''}>
```
redimp: This causes a duplicate DOM id, since it's inserted twice (from `template_html_sidebar_left_inject()` and `template_html_sidebar_right_inject()`) better make this `backlinks-toc-left` and `backlinks-toc-right`.

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
```diff
@@ -86,25 +90,22 @@ def _update_references_for_file(self, filepath):
...
             for match in matches:
                 # Clean up the page name
-                target_page = match.strip()
-
-                # Normalize the page path (handle case sensitivity)
-                if not self.app.config.get("RETAIN_PAGE_NAME_CASE", False):
-                    target_page = target_page.lower()
+                target_page = urllib.parse.unquote(match.strip())
+                target_page = target_page.split("#", 1)[0].strip()
+                target_page = self.normalize_page_name(target_page)
```
redimp, line 109: Add a 
```python
if not target_page: continue
```
here to avoid `[[Title| #section]]` slipping past `strip()` and `split("#")` shich makes it empy

redimp (commented):
Sorry I missed two things and have to request two more changes. Thanks for the work.

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
ribbal: updated

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
ribbal: updated the id to include 'left'/'right'

docs/plugin_examples/plugin_referencingpages/otterwiki_referencingpages.py
ribbal: fixed

redimp (approved):
Looks good to me.

redimp:

PR #529, opened 2026-07-11
move 'on this page' to below 'page index'

### LINKED ISSUE #528: [Change] Move 'On this page' on Left Sidebar
When the 'On this page' list snaps to the left sidebar (_when reducing the browser screen width_), it **shifts the main 'Page index' down**.  This feels a little disjointed.

The page index should remain in its fixed position and the 'On this page' should display below it, which feels a much smoother/natural.

Also, displaying it below the page index would be more consistent with any sidebar plugins (_for an example, see the enhanced referencing pages plugin #527_)

ribbal:
Fixes: #528

### DIFF
--- a/otterwiki/templates/page.html
+++ b/otterwiki/templates/page.html
@@ -15,6 +15,9 @@
 {{ super() }}
 {% include 'snippets/menu.html' %}
 {% include 'snippets/menutree.html' %}
+<div id="sidebar-toc" class="sidebar-toc d-xl-none">
+{% include 'snippets/toc.html' %}
+</div>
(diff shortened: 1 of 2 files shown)


redimp:
Copied the comment from https://github.com/redimp/otterwiki/issues/528#issuecomment-4948991309

Hey @ribbal, thanks for bringing this up and for submitting a PR!

I agree that the shift when resizing is a bit disturbing. But this is a price I'm willing to pay for having the more important information on top: when the "On this page" list is displayed, it refers to the page you are currently reading, so it should be the first thing in the sidebar. The page index is more general and moves down in this case.

PR #530, opened 2026-07-13
Update backlinks when a page is renamed

ribbal:
This change updates the following '_backlinks_' when a page is renamed:

- WikiLinks
- Markdown links
- links to attachments on the renamed page

~~The functionality is controlled by a new UPDATE_LINKS_ON_RENAME config setting (default on)~~

The functionality is controlled by a new **Update backlinks** checkbox on the Rename page (default off) - it retains the users last setting during the current browser session.

The following plugin hooks are implemented:

- `repository_changed` - uses the existing hook, but includes all the pages where backlinks were updated
- `page_saved` - for each of the pages where backlinks were updated

Fixes: a) and b) on #228.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- insertion marker -->
 
+## [v2.22.2](https://github.com/redimp/otterwiki/releases/tag/v2.22.2) - 2026-08-02
+
+<small>[Compare with v2.22.1](https://github.com/redimp/otterwiki/compare/v2.22.1...v2.22.2)</small>
+
+### Security Fixes
+
+- Prevent path traversal via attachment paths and the DataTable embedding `src`
+  option, which allowed reading files outside the wiki repository. Thanks to
+  @NomanProdhan for responsibly disclosing this issue
+  ([62a5336](https://github.com/redimp/otterwiki/commit/62a5336)).
+
 ## [v2.22.1](https://github.com/redimp/otterwiki/releases/tag/v2.22.1) - 2026-07-30
 
 <small>[Compare with v2.22.0](https://github.com/redimp/otterwiki/compare/v2.22.0...v2.22.1)</small>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   @NomanProdhan for responsibly disclosing this issue
   ([62a5336](https://github.com/redimp/otterwiki/commit/62a5336)).
 
+### Dependencies
+
+- Bump gitpython from 3.1.54 to 3.1.55 ([78d6ba6](https://github.com/redimp/otterwiki/commit/78d6ba6)).
+
 ## [v2.22.1](https://github.com/redimp/otterwiki/releases/tag/v2.22.1) - 2026-07-30
 
 <small>[Compare with v2.22.0](https://github.com/redimp/otterwiki/compare/v2.22.0...v2.22.1)</small>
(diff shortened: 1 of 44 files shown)


ribbal:
thanks to @skirmess for their initial work creating a plugin, which formed the basis of implementing this PR into otter wiki core code.

redimp (changes requested):
Wow, thats a big one. Thanks for the work @ribbal (and @skirmess for the plugin).

Some notes from the first glance:

(1) This should not be enabled by default. This is a behavior change that admins have to be aware of. 

(2) Users should be able to pick if they want to update backlinks, please add a checkbox that has to be enabled in the rename form.

(3) This needs tests. Make sure to cover various cases. I stumbled over a bug on first try:

Pages with spaces (or other URL-encoded characters) in the name break Markdown-link rewriting both ways. `[x](/Old%20Page)` yields `yields get_filename("Old%20Page") == "old%20page.md"`

ribbal:
@redimp, I've addressed the points that you raised:

1. The new checkbox (see point 2 below) is unchecked by default

2. Added a new checkbox to the **Rename** page. It is initially unchecked by default, and thereafter defaults to the user's most recent selection during the current browser session

3. Added tests to cover the new behaviour

redimp:
Thanks for the updates, the checkbox and the encoding fix look good.

I pushed two tests to `test_rename_with_updated_links.py` that surface a bug in the current implementation: the rewritten backlinks are only written to the working tree (`storage.update()`) but never staged, and the rename commits with `no_add=True`, so the updates never make it into git history and the repo is left dirty.

- `test_rename_commits_the_rewritten_backlinks`: asserts the rewrite is in the commit (`git show HEAD:...`), not just the working tree.
- `test_rename_leaves_no_uncommitted_changes`: asserts a clean `git status` after the rename.

Both fail as-is. Could you make the rename stage the rewritten pages so they're part of the commit? These should go green once that's fixed.

ribbal:
> Thanks for the updates, the checkbox and the encoding fix look good.
> 
> I pushed two tests to `test_rename_with_updated_links.py` that surface a bug in the current implementation: the rewritten backlinks are only written to the working tree (`storage.update()`) but never staged, and the rename commits with `no_add=True`, so the updates never make it into git history and the repo is left dirty.
> 
> * `test_rename_commits_the_rewritten_backlinks`: asserts the rewrite is in the commit (`git show HEAD:...`), not just the working tree.
> * `test_rename_leaves_no_uncommitted_changes`: asserts a clean `git status` after the rename.
> 
> Both fail as-is. Could you make the rename stage the rewritten pages so they're part of the commit? These should go green once that's fixed.

@redimp, I've applied a fix that should resolve this issue of unstaged files.

redimp:
I keep coming back to reviewing this and always stumble about something ... so I started to fix what I've found. 

The problem I looked at yesterday: Page.rename stages the attachment mv, page mv, and backlink rewrites, then commits them all in one final commit for atomicity. If any step failed after staging (e.g. an error during backlink rewriting), the staged partial rename was left dangling in the index/working tree, leaving the repo dirty.

To address this I addded in:

- `otterwiki/gitstorage.py`: added a reset(revision="HEAD") method that does `git reset --hard` to discard staged + working-tree changes, validating the revision through the existing allowlist and wrapping git errors in StorageError.
- `otterwiki/wiki.py`: wrapped the staging-and-commit block in `Page.rename` in try/except; on any exception it calls `storage.reset()` to roll back to `HEAD`, then re-raises so `handle_rename` still shows the "Renaming failed"  toast. The `page_saved` hooks stay outside the try, so they only fire after a successful commit.
- `tests/test_rename_with_updated_links.py`: added   `test_rename_rolls_back_when_backlink_rewrite_fails`, which monkeypatches rename_backlinks to raise mid-rename and asserts the repo stays clean, the page keeps its old name, the new name doesn't exist, and the backlink content is untouched.


I will look into this again tonight, I want to get this feature finished and out of the door.

redimp (approved):
I reviewed, tested and fixed. I think we are good here.

redimp:
.. and github can not merge it. I will rebase and push this manually.

redimp:
Released in [v2.23.0](https://github.com/redimp/otterwiki/releases/tag/v2.23.0)

PR #534, opened 2026-07-17
Extend Page Index Title

ribbal:
Extend the Title on the Page Index page **from this**:

image

**to this**:

image

### DIFF
--- a/otterwiki/pageindex.py
+++ b/otterwiki/pageindex.py
@@ -72,6 +72,11 @@ class PageIndex:
         else:
             self.path, self.breadcrumbs = None, None
             self.index_depth = 0
+        self.pagename = (
+            get_pagename_for_title(filepath=self.path)
+            if self.path is not None
+            else None
+        )
 
         t_start = timer()
         # get all files in the storage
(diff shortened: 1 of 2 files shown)


redimp:
Looks good to me, will do some tweaks.

ribbal:
@redimp, I would like to make a suggestion related to [commit d131be2](https://github.com/redimp/otterwiki/commit/d131be2ec7cba591fec04f4257ddcf4487f76952).

The commit sets `full=True` in the call to `get_pagename_for_title` to display the **full page path**.  I believe that this could get quite verbose/ugly with long page paths with many folder levels; particularly given the large font used for the page index title.

@deseven and I had a similar conversation regarding 'referencing pages'.  See their [comment dated 29th May](https://github.com/redimp/otterwiki/pull/498#issuecomment-4576002390), which recommends that the full page path should **NOT** be displayed, as the full page path is available by hovering over the referencing page link.

I would suggest that in this case (_Page Index_) it should be consistent and only display the **page name**, and also, the full page path is visible in the breadcrumbs.

redimp:
> I would suggest that in this case (Page Index) it should be consistent and only display the page name, and also, the full page path is visible in the breadcrumbs.

Fair point. The breadcrumbs indicate which Page Index is displayed.

redimp:

PR #539, opened 2026-07-22
Update Styling/Alignment on PageIndex Embedding

ribbal:
Fix styling and alignment on PageIndex embedding to match the standard page index.

**Standard Page Index Example**:

image

**Page Index Embedding Example**:

image

**Issues Fixed**:

1.  the single character column headings have a larger font than the standard page index
2. the columns do not extend to the full page width
3. the standard page index title displays slightly lower than a standard page title

### DIFF
--- a/otterwiki/renderer_embeddings.py
+++ b/otterwiki/renderer_embeddings.py
@@ -1196,7 +1196,7 @@ Options:
                 + '</ul></div>'
             )
 
-        togglediv = f"""<div class="d-inline-block custom-switch font-size-12 mb-10" style="transform: translateY(-36px);float:right;">
+        togglediv = f"""<div class="d-inline-block custom-switch font-size-12 mb-10" style="transform: translateY(-44px);float:right;">
     <input type="checkbox" id="switch-headings" {"checked" if show_toc else ""} value="" onchange="otterwiki.toggleClass(event.target.checked,'pagetoc')"><label for="switch-headings">Toggle page headings</label>
 </div>"""
 
--- a/otterwiki/renderer_embeddings.py
+++ b/otterwiki/renderer_embeddings.py
@@ -1196,7 +1196,7 @@ Options:
                 + '</ul></div>'
             )
 
-        togglediv = f"""<div class="d-inline-block custom-switch font-size-12 mb-10" style="transform: translateY(-44px);float:right;">
+        togglediv = f"""<div class="d-inline-block custom-switch font-size-12 mb-10">
     <input type="checkbox" id="switch-headings" {"checked" if show_toc else ""} value="" onchange="otterwiki.toggleClass(event.target.checked,'pagetoc')"><label for="switch-headings">Toggle page headings</label>
 </div>"""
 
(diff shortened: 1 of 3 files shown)


ribbal:
This shows the updated page index embedding following the fixes:

image

redimp:
Hey @ribbal, thank you so much for taking care of the details!

I will review, test and merge as soon as I find time. I'm currently struggling to keep up. Please bring patience.

redimp (commented):
Thanks for tackling these alignment issues. The three fixes look reasonable and the visual result is an improvement. A few things I'd like to flag before merging:

**1. The toggle only aligns when `{{PageIndex}}` is the first element after the title.**

The previous inline `transform: translateY(-36px); float:right;` positioned the toggle regardless of where the embedding sat. The new CSS depends on DOM adjacency:

```css
article.page > h1 + .custom-switch { ... }
article.page > h1 + .custom-switch + .pageindex-embedding { ... }
```

Rendering `# Title` + intro paragraph + `{{PageIndex}}` produces `h1, p, div.custom-switch, div.pageindex-embedding`. The `<p>` breaks the `h1 + .custom-switch` adjacency, so neither rule matches and the toggle falls back to a full-width inline block with no alignment. Is the "embedding immediately under the title" the only supported layout, or should this handle content between the title and the index too?

**2. Magic negative margins are tied to the current title size.**

`margin-top: -4.4rem` and `-2.75rem` are tuned to the current `.page h1` font-size / line-height. A theme or config that changes the title size will misalign the toggle. Not blocking, just a maintenance note.

**3. The `:before` rule appears to be a no-op.**

```css
.pageindex-embedding:before { content: ""; display: block; }
```

An empty block pseudo-element with no height, margin, or `clear` collapses to zero and has no visible effect (float clearing is already handled by `clear: both` in the adjacency rule). Looks like a leftover that can be removed.

**4. `article.page > h1 + .custom-switch` is broader than the target.**

It matches any `.custom-switch` right after an `h1` in a page article, not just the PageIndex toggle. Low risk, but scoping it to `.pageindex-embedding` would be safer.

Nice work on the font-size fix in particular. The specificity bump (`.page .pageindex-header`) is the right way to override `.page h2` there.

Happy to take this over and address these points myself if that is easier for you, just let me know.

ribbal:
Hi @redimp,

> 1. Is the "embedding immediately under the title" the only supported layout, or should this handle content between the title and the index too?

Regarding point 1 - here is the markdown that I used for those pages above (shown in the screenshots):

```
# Parent

{{PageIndex
|src=*
|style=pageindex
|toc=false
|toggle=true
}}
```

I think that this **should** handle content between the title and the index too - we should allow the user to layout the page as they wish, and the markdown parser should handle it sensibly.

I've just tested this scenario, and I see that the toggle drops back to the left (_instead of being aligned on the right like other page indexes_).  So a little more tweaking required here.

---

> Happy to take this over and address these points myself if that is easier for you

I would be more than happy for you to take this over (_really appreciated_).  I am not familiar with css, but managed to get the page index embedding looking presentable.  If I were to address the issues you have raised, I'm certain they would require further rework/tweaks.

redimp (approved):
Hey @ribbal, nevermind after playing around with it I think of (1) more of as a feature than as a bug. Fixed a minor issue.

redimp:
Released in [v2.22.1](https://github.com/redimp/otterwiki/releases/tag/v2.22.1).

PR #546, opened 2026-08-02
Fix inconsistencies in the readme, add Gateway API HTTPRoute support

nerothank:
The current version of the helm chart provided by _An Otter Wiki_ works well and provides most functionality that most users would require. For my specific use-case there are some things missing, that I would like to contribute. Apart from new functionality there are also some changes to the README  that fix inconsistencies and missing information.

# Proposed changes

**New functionality**
- Add support for GatewayAPI HTTPRoute configuration
- Add value to optionally configure labels to be added to pods
- Adds documentation for the new functionality in the README

**Fixes to the README**
- Mention the option of using `persistence.existingClaim` that was undocumented before, and add missing value table entry
- Add missing values table entries for:
  - `ingress.annotations`
  - `persistence.annotations`
  - `persistence.subPath`

**Common**
- Update the helm chart version to `0.2.0` due to the new functionality introduced.

# Tests

The chart has been successfully tested:
- using only "old" values, assuring the update does not break current installations
- with the new values incorporated, ensuring the new functionality actually works
- with the new default values provided with the chart

### DIFF
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -23,7 +23,9 @@ spec:
       {{- end }}
       labels:
         {{- include "otterwiki.selectorLabels" . | nindent 8 }}
-        {{- toYaml .Values.podLabelsAdditional | nindent 8 }}
+        {{- with .Values.podLabelsAdditional }}
+          {{- toYaml . | nindent 8 }}
+        {{- end }}
     spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -23,6 +23,7 @@ spec:
       {{- end }}
       labels:
         {{- include "otterwiki.selectorLabels" . | nindent 8 }}
+        {{- toYaml .Values.podLabelsAdditional | nindent 8 }}
     spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
--- a/helm/templates/httproute.yaml
+++ b/helm/templates/httproute.yaml
@@ -1,0 +1,38 @@
+{{- if and .Values.httproute.enabled (semverCompare ">=1.25-0" .Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1") -}}
+{{- $fullName := include "otterwiki.fullname" . -}}
+{{- $svcPort := .Values.service.port -}}
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: {{ $fullName }}
+  labels:
+    {{- include "otterwiki.labels" . | nindent 4 }}
+  {{- with .Values.httproute.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  hostnames:
+    {{- toYaml .Values.httproute.hostnames | nindent 4 }}
+  parentRefs:
+    {{- toYaml .Values.httproute.parentRefs | nindent 4 }}
+  rules:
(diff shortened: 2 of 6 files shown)


helm/templates/httproute.yaml
```diff
@@ -0,0 +1,38 @@
...
+  {{- end }}
+spec:
+  hostnames:
+    {{- toYaml .Values.httproute.hostnames | nindent 4 }}
+  parentRefs:
+    {{- toYaml .Values.httproute.parentRefs | nindent 4 }}
+  rules:
+  {{- range .Values.httproute.rules }}
+    - backendRefs:
+        - name: {{ $fulName }}
```
redimp: Typo: `$fulName` so `helm lint` and every `helm install`/`template` fails:

```
Error: parse error at (otterwiki/templates/httproute.yaml:22): undefined variable "$fulName"
```

```suggestion
        - name: {{ $fullName }}
```

helm/templates/httproute.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if and .Values.httproute.enabled (semverCompare ">=1.25-0" .Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1") -}}
```
redimp: `semverCompare` takes exactly 2 args, but here it gets 3 (`">=1.25-0"`, `.Capabilities.APIVersions.Has`, and the API string), so whenever `httproute.enabled=true` this errors:

```
Error: ... wrong number of args for semverCompare: want 2 got 3
```

The `APIVersions.Has` capability check is the correct guard on its own (Gateway API `v1` already implies k8s 1.25+), so the semver check can be dropped:

```suggestion
{{- if and .Values.httproute.enabled (.Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1") -}}
```

helm/templates/deployment.yaml
```diff
@@ -23,6 +23,7 @@ spec:
       {{- end }}
       labels:
         {{- include "otterwiki.selectorLabels" . | nindent 8 }}
+        {{- toYaml .Values.podLabelsAdditional | nindent 8 }}
```
redimp: With the default `podLabelsAdditional: {}`, `toYaml {}` emits a stray `{}` line inside the labels mapping, producing invalid YAML on every default install:

```
Error: YAML parse error on otterwiki/templates/deployment.yaml: ... did not find expected key
```

Guard it with `with` like the other optional blocks in the chart:

```suggestion
        {{- with .Values.podLabelsAdditional }}
        {{- toYaml . | nindent 8 }}
        {{- end }}
```

redimp (changes requested):
Hey @nerothank,

Thanks for the Helm chart improvements! These are genuinely useful additions and
cleanups. Nice work on:

- **Gateway API `HTTPRoute` support** as a first-class alternative to Ingress,
  including the `NOTES.txt` output for it
- **`persistence.existingClaim`** so users can bring their own PVC (plus the requirements note that PV provisioning isn't mandatory)
- **`ingress.annotations` / `httproute.annotations` / `persistence.annotations`** and **`podLabelsAdditional`** for customizing generated resources
- The **`values.yaml` regrouping** into logical sections and the **README table** now documenting the previously-missing options

I did find three template bugs that block rendering (I know the PR is still a draft/untested, flagging them so they're captured). I've left inline comments with suggested fixes on each.

Please review and fix what I've found. Make sure `helm lint` works. Looking
forward to merge this.

nerothank:
Hi @redimp 

Thank you for your valuable insights. I fixed the issues you mentioned and performed first tests on my k3s cluster.
It seems to work well so far. As it is getting late here and I have not tested to the extend I would like, I will still leave this on _draft_ for now.

I should be able to test in more detail this weekend.

PR #556, opened 2026-08-12
Fix broken WikiLink false positives for Mermaid diagram nodes (#550)

### LINKED ISSUE #550: Bug: Broken WikiLinks detected within Mermaid diagrams
The feature to detect broken WikiLinks at https://wiki.example.com/-/housekeeping finds `"Switch"` as broken link, while it is defined like this:

````raw
```mermaid
[...]
        attic_switch[["Switch"]]:::switch
[...]
```
````

### PR DESCRIPTION
**Pull request by agu2347**
## Bug

The housekeeping "Broken WikiLinks" check misreports Mermaid diagram node labels as broken WikiLinks, e.g. `attic_switch[["Switch"]]` gets flagged as a link to a nonexistent page `"Switch"`.

Fixes #550

## Root cause

`handle_housekeeping_brokenwikilinks()` runs `WIKI_LINK_PATTERN` (`\[\[...\]\]`) directly over each page's raw markdown source, including the contents of fenced code blocks. Mermaid's double-bracket "subroutine" node shape (`node_id[["Label"]]`) is syntactically indistinguishable from a WikiLink to that regex, so every such node inside a ` ```mermaid ` fence was reported as a broken link to a page that was never meant to exist.

## Fix

Strip fenced code blocks (both ` ``` ` and `~~~` style, as used for Mermaid diagrams and any other fenced code) from the page content before running `WIKI_LINK_PATTERN` over it. Real WikiLinks outside of code fences are unaffected.

## Testing

Added two tests to `tests/test_housekeeping.py::TestHousekeepingBrokenWikilinks`:

- `test_housekeeping_broken_wikilinks_ignores_mermaid_nodes` -- a page containing only a Mermaid diagram with `[["..."]]` nodes reports no broken links.
- `test_housekeeping_broken_wikilinks_mermaid_does_not_mask_real_links` -- a page with both a Mermaid diagram and a genuinely broken WikiLink still reports the real broken link, proving the fix doesn't mask real diffs.

Verified: `pytest tests/test_housekeeping.py` passes 27/27 (25 existing + 2 new). Full suite: 731 passed, 1 pre-existing failure unrelated to this change (`test_diff_route_output_injection_does_not_corrupt_page`, fails identically on unmodified `main`). Reverting just the `otterwiki/tools.py` change makes both new tests fail with the exact bug from the issue (`"Switch"` reported as a broken link), confirming they're a genuine regression guard. `black --check` (the project's formatter, per `.pre-commit-config.yaml`) is clean on both changed files.

### DIFF
--- a/otterwiki/tools.py
+++ b/otterwiki/tools.py
@@ -159,6 +159,17 @@ def handle_housekeeping_emptypages(form):
     )
 
 
+# Fenced code blocks (```...``` or ~~~...~~~), e.g. mermaid diagrams.
+# Mermaid's double-bracket "subroutine" node shape (e.g. `x[["Switch"]]`)
+# looks exactly like a WikiLink to WIKI_LINK_PATTERN below, so fenced
+# code has to be stripped from the page content before scanning for
+# WikiLinks, or every such node gets misreported as a broken link.
+FENCED_CODE_BLOCK_PATTERN = re.compile(
+    r"^[ \t]{0,3}(`{3,}|~{3,})[^\n]*\n.*?^[ \t]{0,3}\1[ \t]*$",
+    re.DOTALL | re.MULTILINE,
+)
+
+
 def handle_housekeeping_brokenwikilinks(form):
     """Analyze pages for broken WikiLinks."""
     if not has_permission("WRITE"):
@@ -184,7 +195,11 @@ def handle_housekeeping_brokenwikilinks(form):
             continue
         current_pagename = get_pagename(filename, full=True)
 
-        wikilinks = WIKI_LINK_PATTERN.findall(content)
+        # Strip fenced code blocks (e.g. mermaid diagrams) so their
+        # contents are never mistaken for WikiLinks.
+        content_without_code = FENCED_CODE_BLOCK_PATTERN.sub("", content)
+
+        wikilinks = WIKI_LINK_PATTERN.findall(content_without_code)
 
         if not wikilinks:
             continue
--- a/otterwiki/tools.py
+++ b/otterwiki/tools.py
@@ -164,8 +164,22 @@ def handle_housekeeping_emptypages(form):
 # looks exactly like a WikiLink to WIKI_LINK_PATTERN below, so fenced
 # code has to be stripped from the page content before scanning for
 # WikiLinks, or every such node gets misreported as a broken link.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by redimp** [comment]
Hey @agu2347, thank you for this PR! 

I think this approach is right to solve this. 

One change before merging. `FENCED_CODE_BLOCK_PATTERN` uses `\1`, which
requires the closing fence to be exactly as long as the opener and to exist at all. mistune closes on `^ {0,3}<char>{len(opener),}[ \t]*(?:\n|$)` and lets an unclosed fence run to EOF, so we get it wrong both ways:

- Closing fence longer than the opener (```` ``` ```` opened, ````` ```` ````` closed) renders fine today, but the regex misses the close, runs on to the next fence and strips everything in between. Real broken links in that range are swallowed, and nodes from the second block get reported instead.
- Unclosed fence: mistune treats the rest as code, the regex strips nothing, and the original bug still fires.

This matched the renderer on every case I tried:
```python
FENCED_CODE_BLOCK_PATTERN = re.compile(
    r"^ {0,3}(?P<fence>(?P<c>[`~])(?P=c){2,})[^\n]*\n"
    r".*?(?:^ {0,3}(?P=fence)(?P=c)*[ \t]*$|\Z)",
    re.DOTALL | re.MULTILINE,
)
```

`(?P=fence)(?P=c)*` is mistune's length rule, `|\Z` covers the unclosed case. mistune allows spaces only in the indent, not tabs.

agu2347:
Pushed the fix using the regex from your review -- thanks for working that out, it's a much better fix than what I had. Also added two regression tests reproducing both failure modes you described (closing fence longer than the opener, and an unclosed fence running to EOF), confirmed they fail against the old pattern and pass against the new one, and reran the full suite (721 passed, 1 pre-existing unrelated failure). Ready for another look whenever you have time.

redimp:
Released in [v2.24.0](https://github.com/redimp/otterwiki/releases/tag/v2.24.0).

briped:
Inline `[[code]]` should probably also be taken into account :)

redimp:
> Inline `[[code]]` should probably also be taken into account :)

Yes, this covered by https://github.com/redimp/otterwiki/commit/7125fbd1c5851a10b5ecd1e34c077e5e93c14744.

PR #570, opened 2026-09-21
feat: configurable custom 404 page (NOT_FOUND_PAGE), closes #197

### LINKED ISSUE #197: Add ability to customize the 404 page
Currently the 404 page is hardcoded (templates/page404.html). In our case we'd like to customize that page as we're migrating away from an old Trac wiki and wish to let people know why they may be hitting a 404 page in Otterwiki. While we can modify that file directly, it is a rather bad option as it makes Git unhappy. Two options I can think of to make this better:

# Allow defining a custom 404 URL

This would default to an empty string. If it is an empty string, Otterwiki would default to its current behavior. Alternatively there could be a separate boolean to turn this on coupled with a string (HTML) field. As far as I can see, implementing this requires these steps:

* *templates/admin.html* needs a new entry for custom 404 URL setting (empty by default)
* *preferences.py* method *handle_app_preferences()* needs to be modified to commit the new 404 URL setting 
* *wiki.py* method *exists_or_404()* would need to check app.config['404_URL'] and act accordingly

Does this look correct? Is something else required, e.g. for database migrations?

# Allow embedding custom HTML code snippets to page404.html

The implementation might be slightly simpler than above(?). Basically the snippet would be empty by default. If defined, it would override or get prepended/appended the default page404.html.

-----

Any thoughts regarding this? I could take a stab at implementing this myself as it looks deceptively simple to do 😂 .

yukidevv:
Closes #197

## Summary

Adds a configurable custom 404 page. An admin can point `NOT_FOUND_PAGE` to a wiki page (e.g. `Meta/NotFound`), and its rendered content is shown whenever a page is not found. This makes it easy to explain missing pages, e.g. after migrating from another wiki, without modifying `page404.html`.

## Changes

- `server.py`: new setting `NOT_FOUND_PAGE` (default `""`), so it can also be set via `settings.cfg` or an environment variable.
- `admin.html` / `preferences.py`: new "Custom 404 Page" field in Application Preferences, next to "Custom Home Page". Leading/trailing slashes are stripped; paths ending in `.md` and special pages (`/-/...`) are rejected.
- `wiki.py`: `exists_or_404()` renders the configured page with the regular renderer and passes it to `page404.html`.
- `page404.html`: shows the custom content if available, otherwise the default message.

## Design decisions

- **Embedding instead of redirecting**: the response stays a real 404 (better for crawlers and link checkers), the requested path stays visible, and the "Do you want to create it?" link is preserved.
- **Graceful fallback**: if the setting is empty or the configured page does not exist, the default 404 message is shown.
- **No shadowing of the environment**: the preference is stored with `_update_preference_if_changed()`, so saving the form does not override a value set via environment variable.
- No database migration needed, the value is stored in `Preferences`.

## Tests

`tests/test_not_found_page.py` covers the default and custom 404 page, slashes in the setting, a missing configured page, the 404 on `/history`, and saving/rejecting the setting in the admin preferences.

Happy to adjust anything, e.g. switch to a redirect if you prefer that approach.


redimp:
Hey @yukidevv, thanks for this PR.

From the first glance it looks good to me, will test and review in the next days.

